Issue #65, opened 2018-02-23
Make Myokit work with sundials 2.x and 3.x


MichaelClerx:
Find out a way to make Myokit work with sundials 2 and sundials 3 (e.g. the cvode_dense.h include has changed).

So far (feb 2018) Ubuntu and Fedora both ship 2.x (and not the newest releases either)

See e.g. https://chaste.cs.ox.ac.uk/trac/changeset/04c367eed610358fe42f8ba97355ada2b84b131b/git_repo

teosbpl:
Do you have an idea where sundials version string should be placed? It should probably be set automagically by checking library version. Do you know any automagical mechanism for checking version in Python?
It can also be verified by compiling and linking a test source against this library and checking which version fails.
Let me know which version would you prefer.

MichaelClerx:
I think for now an option in myokit.ini (default 2) should be fine. If it becomes common to have 3 installed we can look into doing something more clever...

teosbpl:
SUNLinSolFree and SUNMatDestroy could probably be called on exit to free memory of both objects before destroying object in sim.c and cvode.c (?)

teosbpl:
Verify if this can be closed.

MichaelClerx:
Still need to test this, but will get round to it soon!

Seems [Fedora 28](https://en.wikipedia.org/wiki/Fedora_(operating_system)#Releases) will [ship sundials 3.1.0](https://apps.fedoraproject.org/packages/sundials)

MichaelClerx:
> SUNLinSolFree and SUNMatDestroy could probably be called on exit to free memory of both objects.

Now have this in PR #169

MichaelClerx:
Thanks @teosbpl !!

PR #73, opened 2018-03-01
#65 Sundials 2.x vs 3.0 compatibility via CVDense occurences fixed.

teosbpl:
Compatibility with Sundials 2.x and 3.0 probably achieved.
1. Added field to myokit.ini in section [sundials]
Per analogiam to Chaste I have used extended numeric values:
version = 30000
version = 26000
Version number has  to be set manually.
Default value is 26000
2. Tracked down error related to usage of CVDense, for version>=3.x new syntax is used.
Both versions do build and for each Simulation object is properly created.
3. Thorough tests not perfomed; could not get python all in test directory to work (?)

### DIFF
--- a/myokit/__init__.py
+++ b/myokit/__init__.py
@@ -157,6 +157,8 @@ FORCE_PYSIDE = False
 SUNDIALS_LIB = []
 # Location of the Sundials (CVODE) header files (.h)
 SUNDIALS_INC = []
+# Sundials major version number. Defaults to 26000.
+SUNDIALS_VERSION = 26000
 # Location of the OpenCL shared library objects (.dll or .so)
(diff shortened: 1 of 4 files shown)


teosbpl:
SUNLinSolFree and SUNMatDestroy could probably be called on exit to free memory of both objects.

MichaelClerx:
`Thanks!!

- Could you stick this and future updates in a separate branch please? That way I can test things without merging them into master
- Not sure about the `26000` precision, because the code just works with all of 2.x, or all of 3.x, so not sure we should give users the impression we need to know the version number more accurately?
- Which tests didn't work? Very interested in the output!

MichaelClerx:
This is now included as part of #169

PR #329, opened 2019-05-24
Added some publication examples as unit tests.

MichaelClerx:
Just an idea, not sure if great


MichaelClerx:
Would be nicer to have these as notebooks, listed in the appendix on the examples page.
(Can be shortened where needed for runtime)

Issue #339, opened 2019-06-13
Removed variables keep their bindings / labels


MichaelClerx:

MichaelClerx:
Fixed in #341

PR #341, opened 2019-06-13
Check that binds/labels are removed when a variable is removed

MichaelClerx:
Otherwise they remain in the model's lists of bindings/labels


MichaelClerx:
Closes #339

PR #407, opened 2019-09-18
Removed issue templates

MichaelClerx:


MichaelClerx:
Bit too much hassle for users, didn't like the interface

PR #437, opened 2019-12-03
Trying to fix travis osx build

MichaelClerx:
https://stackoverflow.com/questions/38670295/homebrew-refusing-to-link-openssl

Seems sundials does install openssl

> ==> Installing sundials dependency: openssl@1.1


MichaelClerx:
Works, but takes 20 minutes...

Issue #101, opened 2018-04-07
Redo CellML 1 XML export using CellML 1 model api


MichaelClerx:
Depends on #98

MichaelClerx:
Remember to sort component alphabetically (#18)

MichaelClerx:
Implemented in cellml_1.Model.from_myokit_model()

- [ ] Add stimulus current based on protocol
- [ ] Create components & variables for states, so that they appear in the order of the state variables.
- [ ] Add remaining components & variables
- [ ] Export oxmeta annotations
- [ ] Don't bother with nesting

Issue #434, opened 2019-12-01
Use lxml instead of default elementtree


MichaelClerx:
Advantages

- Line numbers can be shown for import errors
- Better handling of namespaces when exporting
- Pretty xml printing when exporting
- Canonical xml output
- (Almost always much) Faster

Disadvantages

- Extra dependency, one more thing for users to install

MichaelClerx:
Means 1 extra dependency, so think about it!

MichaelClerx:
Currently have the option to use it, but assuming it isn't installed.

MichaelClerx:
Anaconda provides lxml by default

MichaelClerx:
It adds 15-30 seconds to installation time in testing. Probably worth it?

Issue #17, opened 2018-02-23
Stop enforcing unique variable names in CellML export


MichaelClerx:
CellML: During export, don't generate unique variable names (but check depending on the scoping).

MichaelClerx:
This could be a nice-to-have feature once the new API is in place #98

PR #445, opened 2019-12-13
CellML Export via API

MichaelClerx:
See #101


MichaelClerx:
Fixes #17. Fixes #101. Fixes #434

Issue #506, opened 2020-03-21
Capturing breaks when `sys.__stdout__` is None


MichaelClerx:
https://docs.python.org/3/library/sys.html#sys.__stdout__

> Note Under some conditions stdin, stdout and stderr as well as the original values __stdin__, __stdout__ and __stderr__ can be None. It is usually the case for Windows GUI apps that aren’t connected to a console and Python apps started with pythonw.

This seems to be happening for at least one user, leading to

```
  File "C:\ProgramData\Anaconda3\envs\base36\lib\site-packages\myokit\_sim\__init__.py", line 168, in _compile
    with myokit.SubCapture() as s:
  File "C:\ProgramData\Anaconda3\envs\base36\lib\site-packages\myokit\_aux.py", line 182, in __enter__
    self._start_capturing()
  File "C:\ProgramData\Anaconda3\envs\base36\lib\site-packages\myokit\_aux.py", line 301, in _start_capturing
    self._stdout_fd = sys.__stdout__.fileno()
```

MichaelClerx:
Fixed in #507

PR #507, opened 2020-03-21
SubCapture issue on windows/spyder

MichaelClerx:
See #506


MichaelClerx:
Confirmed this resolves the issue

PR #516, opened 2020-04-11
SBML etree

DavAug:
- Minidom has been replaced by ElementTree. 
- Tests have been extended to check the correctness of the myokit model, defined by the SBML file.
- A flag has been introduced to add a time bound variable to the myokit model (not automatically included in the SBML definition, but implicitly assumed)

The tests do not yet include different SBML versions (only level 2, version 3) because that is the version of the Hodgkin Huxley model. I am planning to add a test for each of the supported versions in the next pull request, when we include compartment, species and reaction support. ( The TwoComp.xml will be there for that purpose)


myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,76 @@ def model(self, path):
...
-        x = dom_child(xmodel, 'listOfUnitDefinitions')
+        x = xmodel.find(self.ns['sbml'] + 'listOfUnitDefinitions')
         if x:
-            self._parse_units(model, comp, x)
+            self._parse_units(model, comp, x, self.ns['sbml'])
+
+        # add time as independent variable (not explicit in SBML format)
+        if bind_time:
+            # add and bind time variable to component
+            comp.add_variable('time')
```
MichaelClerx, line 143: What happens if there's already a variable called 'time' ?

myokit/_sim/fiber_tissue.py
```diff
@@ -365,17 +365,20 @@ def _create_backend(self):
 
         # Define libraries
         libs = []
+        flags = []
```
MichaelClerx, line 368: I think I've managed to fix the OpenCL issues on os/x, but in a separate PR. We'll have to find some way to revert the changes to these three files before we can merge this PR

myokit/formats/sbml/_importer.py
```diff
@@ -67,21 +66,28 @@ def _convert_unit(self, unit):
     def info(self):
         return info
 
-    def model(self, path):
-
+    def model(self, path, bind_time=True):
+        """
+        Returns myokit model specified by the SBML fil provided.
```
MichaelClerx, line 71: Maybe some like "Returns a :class:`myokit.Model` based on the SBML file provided.

Also would need some explanation of what `bind_time` does

myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,76 @@ def model(self, path):
...
+            comp.add_variable('time')
+            time = comp.get('time')
+            time.set_binding('time')
+            # set unit and value
+            if 'time' in self.units.keys():
+                unit = self.units['time']
+            else:
+                unit = myokit.units.s
+                log.warn('Unit of time could not be found in file and was by '
+                         + 'default set to seconds.')
```
MichaelClerx, line 152: You can actually remove the + here and it'll still work!

myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,76 @@ def model(self, path):
...
+            self._parse_units(model, comp, x, self.ns['sbml'])
+
+        # add time as independent variable (not explicit in SBML format)
+        if bind_time:
+            # add and bind time variable to component
+            comp.add_variable('time')
+            time = comp.get('time')
+            time.set_binding('time')
+            # set unit and value
+            if 'time' in self.units.keys():
```
MichaelClerx, line 147: A slightly more efficient way to do this is:
```
try:
    unit = self.units['time']
except KeyError:
    unit = myokit.units.s
    log.warn(...)
```
This version only has a single dict lookup (and some python fans prefer it for more philosophical reasons too)

myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,76 @@ def model(self, path):
         refs = {}
 
         # Handle notes, if given
-        x = dom_child(xmodel, 'notes')
+        x = xmodel.find(self.ns['sbml'] + 'notes')
```
MichaelClerx, line 105: This won't detect if there's two 'notes' elements, but I suppose we're not super interested in detecting invalid files here...

myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,76 @@ def model(self, path):
...
+        x = xmodel.find(self.ns['sbml'] + 'listOfFunctionDefinitions')
         if x:   # pragma: no cover
             log.warn('Custom math functions are not (yet) implemented.')
 
         # Parse custom units
-        x = dom_child(xmodel, 'listOfUnitDefinitions')
+        x = xmodel.find(self.ns['sbml'] + 'listOfUnitDefinitions')
         if x:
-            self._parse_units(model, comp, x)
+            self._parse_units(model, comp, x, self.ns['sbml'])
```
MichaelClerx, line 138: Given that it's a property, I guess we don't strictly need to pass `self.ns['sbml']` here?

myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,76 @@ def model(self, path):
...
         if x:
-            self._parse_rules(model, comp, refs, x)
+            self._parse_rules(model, comp, refs, x, self.ns['sbml'])
 
         # Parse extra initial assignments
-        x = dom_child(xmodel, 'listOfInitialAssignments')
+        x = xmodel.find(self.ns['sbml'] + 'listOfInitialAssignments')
         if x:
-            self._parse_initial_assignments(model, comp, refs, x)
+            self._parse_initial_assignments(model,
```
MichaelClerx, line 169: You can also just write
```
self._parse_initial_assignments(
    model, comp, refs, x, self.ns['sbml'])
```
I find that's usually more readable :D

myokit/formats/sbml/_importer.py
```diff
@@ -159,42 +188,101 @@ def model(self, path):
         # Return finished model
         return model
 
-    def _parse_initial_assignments(self, model, comp, refs, node):
+    def _get_namespaces(self, sbml_version: str, log):
```
MichaelClerx, line 191: Shouldn't we be getting the namespace from the file, instead of getting it from the version attributes?
Or maybe just check that the file combines the two correctly and complain if it says it's version 4 but uses ns for version 3?

myokit/formats/sbml/_importer.py
```diff
@@ -159,42 +188,101 @@ def model(self, path):
...
+        """
+        Defines namespaces for supported SBML versions. Supported versions are:
+        level 2, version 3, 4, 5.
+        """
+        ns = dict()
+        if sbml_version == "{http://www.sbml.org/sbml/level2/version3}":
+            # SBML
+            ns['sbml'] = "{http://www.sbml.org/sbml/level2/version3}"
+            # MathML
+            ns['mathml'] = "{http://www.w3.org/1998/Math/MathML}"
```
MichaelClerx, line 201: Does this change in past or future versions? Otherwise can just set once for all versions, I think

myokit/formats/sbml/_importer.py
```diff
@@ -159,42 +188,101 @@ def model(self, path):
...
+                     % sbml_version + ' may not be imported correctly.'
+                     )
+
+        return ns
+
+    def _get_sbml_version(self, root):
+        """
+        Returns the SBML version of the file.
+        """
+        m = re.match(r'\{.*\}', root.tag)
```
MichaelClerx, line 239: What's happening here? Can we not get the version from either the namespace of the root element, or from the attributes of the root element?

myokit/formats/sbml/_importer.py
```diff
@@ -159,42 +188,101 @@ def model(self, path):
...
-
-    def _parse_parameters(self, model, comp, refs, node):
+    def _parse_parameters(self, model, comp, refs, node, ns):
         """
         Parses parameters
         """
-        node = dom_child(node, 'parameter')
-        while node:
+        nodes = node.findall(ns + 'parameter')
+        for node in nodes:
```
MichaelClerx, line 283: Since `nodes` is never used again, you could also just write `for node in node.findall(..)

myokit/formats/sbml/_importer.py
```diff
@@ -203,89 +291,102 @@ def _parse_parameters(self, model, comp, refs, node):
             else:
                 # Create variable
                 unit = None
-                if node.hasAttribute('units'):
-                    foreign_unit = node.getAttribute('units')
+                if node.get('units'):
```
MichaelClerx, line 294: Not sure you need two lookups here!

MichaelClerx: (The old minidom syntax didn't give a nicer way to do this)

myokit/formats/sbml/_importer.py
```diff
@@ -203,89 +291,102 @@ def _parse_parameters(self, model, comp, refs, node):
...
-            node = dom_next(node, 'assignmentRule')
 
         # Create variables with rate rules (states)
-        node = dom_child(parent, 'rateRule')
-        while node:
+        nodes = parent.findall(ns + 'rateRule')
+        for node in nodes:
             var = self._convert_name(
-                str(node.getAttribute('variable')).strip())
+                str(node.get('variable')).strip())
```
MichaelClerx, line 343: What happens if 'variable' doesn't exit?

myokit/formats/sbml/_importer.py
```diff
@@ -203,89 +291,102 @@ def _parse_parameters(self, model, comp, refs, node):
...
-            while node2:
-                kind = str(node2.getAttribute('kind')).strip()
+            units = node.find(ns + 'listOfUnits')
+            units = units.findall(ns + 'unit')
+            for node2 in units:
+                kind = str(node2.get('kind')).strip()
                 u2 = self._convert_unit(kind)
-                if node2.hasAttribute('multiplier'):
-                    m = float(node2.getAttribute('multiplier'))
+                if node2.get('multiplier'):
```
MichaelClerx, line 379: This looks like it could be more efficient?
What does node2.get('multiplier') do if multiplier isn't set?

myokit/tests/data/formats/sbml/TwoComp.xml
```diff
@@ -0,0 +1,189 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MichaelClerx, line 1: If we're not using this one yet, I'd prefer it didn't come in until a future PR!

myokit/tests/test_formats_importers.py
```diff
@@ -50,6 +50,12 @@ def test_unknown(self):
 
 class SBMLTest(unittest.TestCase):
     """ Test SBML import. """
+    # Test Case I: Hodgkin Huxley
+    i = formats.importer('sbml')
+    HoHu = i.model(os.path.join(
```
MichaelClerx, line 55: Could we call this something in camel case?

myokit/tests/test_formats_importers.py
```diff
@@ -50,6 +50,12 @@ def test_unknown(self):
...
+        DIR_FORMATS,
+        'sbml',
+        'HodgkinHuxley.xml'))
```
MichaelClerx, line 58: Also, there's a `setUpClass` method you can use to initialise a unit test, instead of declaring class properties like this. See e.g. https://github.com/MichaelClerx/myokit/blob/master/myokit/tests/test_simulation_cvode.py#L34

MichaelClerx: We can make multiple Test Classes for multiple models later, if you like

myokit/tests/test_formats_importers.py
```diff
@@ -82,6 +88,145 @@ def sbml(fname):
...
+            'beta_m',
+            'alpha_h',
+            'beta_h',
+            'alpha_n',
+            'beta_n'
+        ]
+
+        # test whether parameters are in myokit model
+        for param in parameters:
+            assert self.HoHu.has_variable('sbml.' + param)
```
MichaelClerx, line 126: I guess you've been using `pytest` :D 

In Myokit we're using plain `unittest` https://docs.python.org/3/library/unittest.html so this becomes `self.assertTrue(self.model.has_variable(..))`

myokit/tests/test_formats_importers.py
```diff
@@ -82,6 +88,145 @@ def sbml(fname):
...
+            'alpha_h': None,
+            'beta_h': None,
+            'alpha_n': None,
+            'beta_n': None
+        }
+
+        # test whether parameters have correct units
+        for param in param_unit_dict:
+            unit = param_unit_dict[param]
+            assert unit == self.HoHu.get('sbml.' + param).unit()
```
MichaelClerx, line 164: self.assertEqual(

myokit/tests/test_formats_importers.py
```diff
@@ -82,6 +88,145 @@ def sbml(fname):
...
+            'g_L': 0.3,
+            'E_Na': -190,
+            'E_K': -63,
+            'E_L': -85.613
+        }
+
+        # test whether parameters have correct initial values
+        for param in param_value_dict:
+            value = param_value_dict[param]
+            assert value == self.HoHu.get('sbml.' + param).value()
```
MichaelClerx, line 184: One downside of doing it with lists and dicts is that if it fails you don't get nice output saying which entry fails, it'll just tell you line 184 failed, but that's not top priority for now :-)

MichaelClerx:
One thing I'd suggest you start with:

1. undo the changes to the OpenCL files in this branch
2. `git merge develop` to get the new fixes for OpenCL on os/x

myokit/formats/sbml/_importer.py
MichaelClerx, line 191: Myokit is used on Python 2 and Python 3, so `sbml_version : str` will have to become just `sbml_version`, I'm afraid

DavAug:
I pulled the latest version of the develop branch, reverted the changes in the opencl.py, openclsim.py and fiber_tissue.py file and then merged the develop branch into sbml-etree. This reintroduced the failures of the tests again, see example below.

Was this expected?

```
======================================================================
ERROR: test_simulation_p (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_simulation_p
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/tests/test_simulation_p.py", line 17, in <module>
    from shared import DIR_DATA, CancellingReporter
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/tests/shared.py", line 33, in <module>
    OpenCL_FOUND = myokit.OpenCL.supported()
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/opencl.py", line 206, in supported
    OpenCL._get_instance()
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/opencl.py", line 90, in _get_instance
    OpenCL()
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/opencl.py", line 65, in __init__
    mname, fname, args, libs, libd, incd)
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/__init__.py", line 197, in _compile
    return load_module(name, d_modul)
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/__init__.py", line 54, in load_module
    module = importlib.util.module_from_spec(spec)
ImportError: dlopen(/var/folders/8w/b9lb3wb92c1fscyr5tc8nswh0000gp/T/tmphwfhy8yimyokit/module/myokit_opencl_info_36.cpython-37m-darwin.so, 2): Symbol not found: _clGetDeviceIDs
  Referenced from: /var/folders/8w/b9lb3wb92c1fscyr5tc8nswh0000gp/T/tmphwfhy8yimyokit/module/myokit_opencl_info_36.cpython-37m-darwin.so
  Expected in: flat namespace
 in /var/folders/8w/b9lb3wb92c1fscyr5tc8nswh0000gp/T/tmphwfhy8yimyokit/module/myokit_opencl_info_36.cpython-37m-darwin.so


----------------------------------------------------------------------
```

myokit/formats/sbml/_importer.py
DavAug: I was planning on implementing the time read in properly in the next pull request. Apparently it has changed how the SBML versions deal with time, and dependent on the version they assume different base units and different identifiers. 

Should I then remove the time binding for now and incorporate and introduce it with the next pull request?

myokit/_sim/fiber_tissue.py
DavAug: reverted and merged the develop branch, bit this didn't fix the bug, see comment below.

MichaelClerx:
@DavAug I accidentally sent a fix to `master` instead of `develop` :man_facepalming: 

Have stuck the fix on develop now too, could you try merging in develop again for me? Sorry!

myokit/formats/sbml/_importer.py
DavAug: Will do, once we decided what to do with the time binding, see comment above.

myokit/formats/sbml/_importer.py
DavAug: Done

myokit/formats/sbml/_importer.py
DavAug: Is also more elegant. Done.

myokit/formats/sbml/_importer.py
DavAug: Should we push this issue to the next pull request? I will need to read more closely through the SBML specifications anyway to deal with time, compartments, species etc. and then I can make sure that the file obeys the prescribed format.

myokit/formats/sbml/_importer.py
DavAug: Agreed. ns wasn't a property before and I forgot to change the input. Done.

myokit/formats/sbml/_importer.py
DavAug: Ok. Done.

myokit/formats/sbml/_importer.py
DavAug: This won't change, but other properties such as base unit definitions, and other namespaces. For now we could, however, just add the 'mathml' namespace irrespective of the sbml version

DavAug: Done.

myokit/formats/sbml/_importer.py
DavAug: There doesn't seem to be a native function in etree to access the namespace of any variable. So what this does is it gets the tag of the root, i.e. the name/ id, and straps off it's namespace, which is the SBML version.

I haven't found a better way to do this.

myokit/formats/sbml/_importer.py
DavAug: Agreed, and changed everywhere.

myokit/formats/sbml/_importer.py
DavAug: Agreed. Done.

myokit/formats/sbml/_importer.py
DavAug: node.get() returns None if multiplier does not exist, so it wouldn't do anything.

DavAug: I'm not 100 percent sure how myokit units behave, but would something like this produce the same 
```
kind = str(node2.get('kind')).strip()
u2 = self._convert_unit(kind)
if node2.get('multiplier'):
      u2 *= float(node2.get('multiplier'))
 if node2.get('scale'):
       u2 *= 10 ** float(node2.get('scale'))
 if node2.get('exponent'):
        u2 **= float(node2.get('exponent'))
 unit *= u2
```

And what happens anyway if a unit has several `kind`. I wanted to clarify these questions for the next pull request and just kept your structure and simply replaced minidom by etree everywhere.

myokit/tests/data/formats/sbml/TwoComp.xml
DavAug: Ok, I removed it.

myokit/tests/test_formats_importers.py
DavAug: Yep, done.

myokit/tests/test_formats_importers.py
DavAug: Ok, sounds good! I used the setUpClass method now.

myokit/tests/test_formats_importers.py
DavAug: Ok, changed.

myokit/tests/test_formats_importers.py
DavAug: Done

myokit/tests/test_formats_importers.py
DavAug: Yes, that's something I had to deal with a couple times now. How would you do it then? write out the asserts for each parameter explicitly? I'm happy to adapt for the next pul request.

myokit/formats/sbml/_importer.py
DavAug: At this point it is assumed that the file adheres to the SBML file format. It would make sense to have a rateRule without defining the lhs, i.e. the variable. But as mentioned above, I basically just kept your code and replaced mindom by etree. 

I will look thorugh the exact specifications in the next PR and make sure that only files following those will be read in, if that's ok.

myokit/formats/sbml/_importer.py
DavAug: I guess that would make sense! I have, however, not found a native function in etree to access the namespaces in a file. What we could do is look at the SBML definitions, where namespaces should be expected to occur in the file, and then look at all those places. But we probably would have to start branching out from the root, because I don't global searches are possible.

What would you prefer?

DavAug:
Haha ok thats strange. Your develop branch passes but when I merge it into my branch it fails?

DavAug:
Also locally on my OS the errors remained the same.

```
======================================================================
ERROR: test_simulation_p (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_simulation_p
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/tests/test_simulation_p.py", line 17, in <module>
    from shared import DIR_DATA, CancellingReporter
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/tests/shared.py", line 33, in <module>
    OpenCL_FOUND = myokit.OpenCL.supported()
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/opencl.py", line 209, in supported
    OpenCL._get_instance()
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/opencl.py", line 93, in _get_instance
    OpenCL()
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/opencl.py", line 68, in __init__
    mname, fname, args, libs, libd, incd, larg=flags)
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/__init__.py", line 199, in _compile
    return load_module(name, d_modul)
  File "/Users/david/Workspace/sabs-r3/myokit/myokit/_sim/__init__.py", line 54, in load_module
    module = importlib.util.module_from_spec(spec)
ImportError: dlopen(/var/folders/8w/b9lb3wb92c1fscyr5tc8nswh0000gp/T/tmpryt3u3iumyokit/module/myokit_opencl_info_36.cpython-37m-darwin.so, 2): Symbol not found: _clGetDeviceIDs
  Referenced from: /var/folders/8w/b9lb3wb92c1fscyr5tc8nswh0000gp/T/tmpryt3u3iumyokit/module/myokit_opencl_info_36.cpython-37m-darwin.so
  Expected in: flat namespace
 in /var/folders/8w/b9lb3wb92c1fscyr5tc8nswh0000gp/T/tmpryt3u3iumyokit/module/myokit_opencl_info_36.cpython-37m-darwin.so


----------------------------------------------------------------------
Ran 391 tests in 5.650s

FAILED (errors=36)
FAILED
```

DavAug:
I just doubled checked. Also the develop branch does not pass the tests on my machine.

MichaelClerx:
Hmmmm, that's a shame. Well, we can ignore the OpenCL errors for now, in this branch I guess

myokit/formats/sbml/_importer.py
MichaelClerx: No it's fine to keep it in there, I just think we should have some strategy in case the model contains a variable called "time", because that will cause this line to fail. There's an `add_variable_allow_renaming(name)` method you could use instead, for example?

myokit/formats/sbml/_importer.py
MichaelClerx: Yeah it's fine. I'm happy with an importer that just works most of the time, for valid files. There's plenty of validation tools etc. out there already without us adding one :-)

myokit/formats/sbml/_importer.py
MichaelClerx: Oh ok! I see what this is doing now.
There's a method `mxml.split` that splits off the namespace in much the same way. Might be worth using that instead. (Not "better", just means we're using a single method consistently and can fix bugs more easily etc)

myokit/formats/sbml/_importer.py
MichaelClerx: Cool!

That's ok as long as we "fail gracefully".
For example, I think `get()` returns `None` if an attribute doesn't exist, so if we do `str(get(..))` we'd get a string 'None' which might not be what we want

myokit/formats/sbml/_importer.py
MichaelClerx: Fair enough. I'm just not sure if get() is an expensive operation. Probably not?
I was thinking `x = get(); if x: ...` would give us a single get
Up to you :D

myokit/formats/sbml/_importer.py
MichaelClerx: I think you just need to get it from each element separately, so like you're doing with the regex or using `mxml.split()`.
So I could write
```
<model xmlns="namespace">
  <component />
</model>
```
or
```
<sbml:model xmlns:sbml="namespace">
  <sbml:component />
</sbml:model>
```
or
```
<model xmlns="other namespace">
  <component xmlns="namespace" />
</model>
```
and in all of those the `component` tag would be a "{namespace}component".

So no file-wide namespaces, each element has one!

myokit/tests/test_formats_importers.py
MichaelClerx: Exactly. It's not super important but the best way is to write explicit asserts I'm afraid. Maybe I should consider switching to pytest :D

myokit/formats/sbml/_importer.py
```diff
@@ -96,53 +102,72 @@ def model(self, path):
         refs = {}
 
         # Handle notes, if given
-        x = dom_child(xmodel, 'notes')
+        x = xmodel.find(self.ns['sbml'] + 'notes')
         if x:
             log.log('Converting <model> notes to ascii')
-            model.meta['desc'] = html2ascii(x.toxml(), width=75)
+            model.meta['desc'] = html2ascii(ET.tostring(x, encoding='unicode'),
```
MichaelClerx, line 108: I'm guessing that should be encoding='utf-8' ?

myokit/formats/sbml/_importer.py
MichaelClerx: Actually, are there default options specified for these attributes? `.get()` has a `default` option, so you might be able to do something like
`u2 *= float(node2.get('multiplier', default=1))`
etc.

MichaelClerx:
Thanks @DavAug ! I think once you changed the `encoding` it should pass in python 2.7 as well and then we're pretty much done!

myokit/formats/sbml/_importer.py
DavAug: I chose unicode because any other encoding returns a bit string which then has to be converted back. But I can change that, no prob.

MichaelClerx: It seems to be failing in Python 2.7

MichaelClerx: utf-8 should be unicode too, so not sure what's going on there

myokit/formats/sbml/_importer.py
DavAug: OK. Done.

myokit/formats/sbml/_importer.py
DavAug: Done.

myokit/formats/sbml/_importer.py
DavAug: I am using split to access the namespace. I would keep the namespace specification in the beginning for now and the maths ns for the mathy bits for now, as otherwise I will have to prepare a comprehensive list of all the namespaces defined within an SBML version and I would rather do that for the next PR, where the different namespaces may have more relevance. At the moment it seems, we are mostly interested in the maths namespace, so we can translate the equations properly and the gobal version ns to find the relevant nodes.

myokit/formats/sbml/_importer.py
DavAug: Done

myokit/formats/sbml/_importer.py
DavAug: Done

myokit/formats/sbml/_importer.py
DavAug: Done

myokit/tests/test_formats_importers.py
DavAug: Ok I will leave it for now, and change in the next PR

myokit/formats/sbml/_importer.py
```diff
@@ -67,21 +66,34 @@ def _convert_unit(self, unit):
     def info(self):
         return info
 
-    def model(self, path):
+    def model(self, path, bind_time=True):
+        """
+        Returns a :class:myokit.Model based on the SBML file provided.
 
+        Arguments:
+            path -- Path to SBML file.
```
MichaelClerx, line 74: Please see e.g. [https://github.com/MichaelClerx/myokit/blob/master/myokit/_model_api.py#L319](here) for the syntax we're using in Myokit.

PR #528, opened 2020-04-25
Extend sbml support

DavAug:
This pull request extends the SBML support of Myokit. Supported features are now

- UnitDefinitions
- Compartments
- Species
- Parameters
- Reactions
- InitialAssignments
- AssignmentRules
- RateRules

Not supported features are:
- FunctionDefinitions - should be doable, but I didn't see an immediate need.
- AlgebraicRules
- Constraints - I don't think constraints can be handled by myokit, can they?
- Events - Seems too complex and unnecessary. For my application the protocol will typically not be in the SBML file anyway.


DavAug:
I am still in the process of writing the tests

DavAug:
I'm tempted to start with a random model from the [SBML test suite](http://sbml.org/Software/SBML_Test_Suite) and then iteratively select more model from the test suite to get 100% coverage. I would then for each model create a separate test class. Is this how you imagined the testing?

myokit/formats/sbml/_importer.py
```diff
@@ -371,39 +1056,43 @@ class SBMLError(myokit.ImportError):
     """
 
 
-unit_map = {
-    'dimensionless': myokit.units.dimensionless,
+# SBML base units according to libSBML (order matters for unit identification!)
```
MichaelClerx, line 1059: How does order matter for identification?

myokit/formats/mathml/_parser.py
```diff
@@ -947,3 +951,18 @@ def _parse_name(self, element):
         except Exception as e:
             raise MathMLError('Unable to create Name: ' + str(e), element)
 
+    def _parse_symbol(self, element):
+        """
+        Parses a ``<csymbol>`` element and returns a :class:`myokit.Name`
+        created by the name factory.
+        """
+
+        timeUri = 'http://www.sbml.org/sbml/symbols/time'
```
MichaelClerx, line 960: We'll need to find some way to not have sbml (or cellml) specific stuff in the mathml parser

MichaelClerx:
> I'm tempted to start with a random model from the [SBML test suite](http://sbml.org/Software/SBML_Test_Suite) and then iteratively select more model from the test suite to get 100% coverage. I would then for each model create a separate test class. Is this how you imagined the testing?

Not exactly :D

Couple of guidelines:

- tests should be short,and test a single "unit", whatever that is
- often, that means 1 test per method, but there's a second important idea: tests should run via the public API. The idea here is that we don't care how something is implemented, we just want to know that the code does what it promises to do, correctly. Ideally, you'd decide _before_ coding starts, what needs to happen, and agree on some kind of API, and then someone will write tests for this, and _after_ that (or in parallel), someone else will write the actual code.
  - In this case a "unit" is probably more something like "parsing units", or "parsing a species definition". You can write quick tests for these without having to create models for each if you do something like the "wrap" method used here: https://github.com/MichaelClerx/myokit/blob/master/myokit/tests/test_cellml_v1_parser.py
- things like coverage testing and flake8 are there to help, don't write tests with the goal making these annoying tools shut up :D . It's probably possible to create some minimal set of tests that will give you full coverage, but then if you need to make a change to the code later, it can be very hard to find which part of the tests actually tests that code. It's much easier if you need to make a change to, say "parse_unit", if there's a "test_unit_parsing" method somewhere

myokit/formats/sbml/_importer.py
```diff
@@ -37,332 +38,1016 @@ def __init__(self):
...
 
         Arguments:
-
-        ``path``
-            The path to the SBML file.
-        `` bind_time``
-            If set to ``True`` (default), a variable called "time" will be
-            created and bound to `time`.
-
+            path -- Path to SBML file.
```
MichaelClerx, line 49: Please update to red code above, to match the style used in the rest of Myokit :-)

myokit/formats/sbml/_importer.py
```diff
@@ -37,332 +38,1016 @@ def __init__(self):
...
+            './'
+            + '{http://www.sbml.org/sbml/level3/version2/core}'
+            + 'listOfParameters/'
+            + '{http://www.sbml.org/sbml/level3/version2/core}'
+            + 'parameter')
+        if params:
+            return params
+        return None
+
+    def _getListOfSpecies(self, element):
```
MichaelClerx, line 800: How do these work? Can there be more than one "listOfSpecies" in the document?

MichaelClerx:
Hey @DavAug ! This looks really cool. I will have to find some time this week to review this carefully, which will probably involve learning a bit about SBML :sweat_smile:

DavAug:
Ok thanks!

myokit/formats/sbml/_importer.py
DavAug: I forgot to remove that comment. It mattered before, because with libSBML the units were numbered and I used an DictKey array to access the keys. 
But now it doesn't matter anymore, because we are using keys directly.

myokit/formats/sbml/_importer.py
DavAug: No, I don't think so. As far as, I understood are models structured like this

beginning of model definition
list of function definitions (optional) (Section 4.3)
list of unit definitions (optional) (Section 4.4)
list of compartments (optional) (Section 4.5)
list of species (optional) (Section 4.6)
list of parameters (optional) (Section 4.7)
list of initial assignments (optional) (Section 4.8)
list of rules (optional) (Section 4.9)
list of constraints (optional) (Section 4.10)
list of reactions (optional) (Section 4.11)
list of events (optional) (Section 4.12)
end of model definition

where each list only appears once (although I don't remember reading this explicitly somewhere).

But since each list has an arbitrary number of elements, i.e.

listOfSpecies
     species
     species
listOfSpecies

there wouldn't be a need to have multiple list objects.

DavAug:
@MichaelClerx Thank you. Let me know if you have any questions. I tried to follow the SBML guidelines as close as possible.

DavAug:
Hi @MichaelClerx 
Never mind about the tests that I wrote, I forgot that you told me about the wrapping of the xml files. I'll do them according to that.

MichaelClerx:
Thanks @DavAug ! Looks like I might have some emergency last-minute work on my hands the next few days, so review will most likely be delayed until next week...

DavAug:
Don't worry and thank you.

DavAug:
There are some tests for errors missing and I also wanted to test the handling of units more explicitly, but before I do that I thought it may be best to wait for you to agree on the tests that I've written so far.

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
-        self.units = {}
 
-    def _convert_name(self, name):
+    def info(self):
+        return info
+
+    def supports_model(self):
         """
-        Converts a name to something acceptable to myokit.
+        Returns a bool indicating if model import is supported.
```
MichaelClerx, line 45: In PINTS we've started doing (most) docstrings for inherited methods as `"""See :meth:`myokit.format.Importer.supports_model()`."""`.
Think that might be useful here

myokit/formats/sbml/_importer.py
```diff
@@ -371,39 +1103,43 @@ class SBMLError(myokit.ImportError):
...
+    'dimensionless': myokit.units.dimensionless,
+    'farad': myokit.units.F,
+    'gram': myokit.units.g,
+    'gray': myokit.units.Gy,
+    'henry': myokit.units.H,
     'hertz': myokit.units.Hz,
-    'litre': myokit.units.L,
-    'newton': myokit.units.N,
-    'sievert': myokit.units.Sv,
+    'item': None,  # TODO: myokit equivalent not yet looked up
```
MichaelClerx, line 1120: What kind of unit is `item` ???

myokit/formats/sbml/_importer.py
```diff
@@ -371,39 +1103,43 @@ class SBMLError(myokit.ImportError):
...
+    'radian': myokit.units.rad,
+    'second': myokit.units.s,
+    'siemens': myokit.units.S,
+    'sievert': myokit.units.Sv,
     'steradian': myokit.units.sr,
+    'tesla': myokit.units.T,
+    'volt': myokit.units.V,
+    'watt': myokit.units.W,
+    'weber': myokit.units.Wb,
+    'invalid': None,
```
MichaelClerx, line 1144: 14 invalids? doesn't look like a unit to me :D

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+        Parses the SBML file at ``path`` and returns a myokit
+        model.
+        """
+        # Read file
+        try:
+            tree = ET.parse(path)
+        except Exception as e:
+            raise SBMLError('Unable to parse XML: ' + str(e))
 
-    def model(self, path, bind_time=True):
```
MichaelClerx, line 69: The `model` method is part of `myokit.formats.Importer`, and is used to provide a unified interface to create Myokit models. E.g. if you implement a `model` method users can type `python myokit sbml import modelmodelmodel.sbml new_model.mmt` to do imports from the command line

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
-        if not self.re_name.match(name):
-            org_name = name
-            name = self.re_white.sub('_', name)
-            name = self.re_alpha.sub('_', name)
-            if not self.re_name.match(name):
-                name = 'x_' + name
-            self.logger().warn(
-                'Converting name <' + org_name + '> to <' + name + '>.')
-        return name
+        return True
```
MichaelClerx, line 47: Should be ``True``, but please implement the `model` method! (see below)

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
 
-    def _convert_unit(self, unit):
+    def supports_component(self):
```
MichaelClerx, line 49: This should only return `True` if you've implemented a `component()` method that adds a component to an existing myokit model by importing it from a file.

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
-            return self.units[unit]
-        elif unit in unit_map:
-            return unit_map[unit]
-        else:      # pragma: no cover
-            raise SBMLError('Unit not recognized: ' + str(unit))
+        return True
 
-    def info(self):
-        return info
+    def parse_file(self, path):
```
MichaelClerx, line 55: I like these methods, but it might be a good idea to have the following structure instead:

```
myokit.formats.sbml
   .SBMLParser
      .parse_string()
      etc
      does dirty work, gives the user fancy options like parsing strings
    .SBMLImporter
      .model()
      .supports_model()
      . etc. This thing just makes calls to an SBMLParser object, and doesn't give all the nice features, but instead provides the unified myokit.formats.Importer interface

File: myokit/formats/sbml/_importer.py [hunk 12]
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
-
-        # Get model node
-        xmodel = root[0]
-        if xmodel.get('name'):
-            name = str(xmodel.get('name'))
-        elif xmodel.get('id'):
-            name = str(xmodel.get('id'))
-        else:
+        # Check whether file has SBML 3.2 namespace
+        ns = self._getNamespace(root)
```
MichaelClerx, line 103: Myokit uses `CamelCase` for classes, and `whats_apparently_called_snake_case` for module, method and variable names.

(With the exception being publically visible module-level constants, which are `WRITTEN_LIKE_THIS_FOR_EXTRA_UGLINESS`, but I don't think you have any of those)

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+        # another PR)
+        funcDefs = self._getListOfFunctionDefinitions(SBMLmodel)
+        if funcDefs:
+            raise SBMLError(
+                'Myokit does not support functionDefinitions. Please insert '
+                'your function wherever it occurs in yout SBML file and delete'
+                ' the functionDefiniton in the file.')
+
+        # Create user defined unit reference
+        self.userUnitDict = dict()
```
MichaelClerx, line 143: Is this meant to be accessed by the user? (And if so, is that safe?) Otherwise make private

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+        for unitId in units:
+            unit = units[unitId]
+            if unit in self.userUnitDict:
+                self.userUnitDict[unitId] = self.userUnitDict[unit]
+            elif unit in SBML2MyoKitUnitDict:
+                self.userUnitDict[unitId] = SBML2MyoKitUnitDict[unit]
+            else:
+                self.userUnitDict[unitId] = None
+
+        # Initialise parameter and species dictionary
```
MichaelClerx, line 175: I assume the contents of these dicts are quite specific? As in, they need to have some specific structure?
In that case, please either add a bunch of comments explaining this structure (could be as simple as "maps ids to myokit.Component objects"), or alternatively create some class structure for this information to live in

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+        self.paramAndSpeciesDict = dict()
+
+        # Add compartments to model
+        compDict = dict()
+        comps = self._getListOfCompartments(SBMLmodel)
+        if comps:
+            for comp in comps:
+                idx = comp.get('id')
```
MichaelClerx, line 183: Would this code be better off in a `_parse_component` method?

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+        units = {
+            'substanceUnit': SBMLmodel.get('substanceUnits'),
+            'timeUnit': SBMLmodel.get('timeUnits'),
+            'volumeUnit': SBMLmodel.get('volumeUnits'),
+            'areaUnit': SBMLmodel.get('areaUnits'),
+            'lengthUnit': SBMLmodel.get('lengthUnits'),
+            'extentUnit': SBMLmodel.get('extentUnits'),
+        }
+        for unitId in units:
+            unit = units[unitId]
```
MichaelClerx, line 167: `for unit_id, unit in units.items():`

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+
+        name = self._convert_name('myokit')
+        if 'MyoKit' in compDict:
+            raise SBMLError(
+                'The compartment ID <MyoKit> is reserved in a myokit'
+                ' import.')
+        compDict['MyoKit'] = model.add_component(name)
+
+        # Add parameters to model
+        params = self._getListOfParameters(SBMLmodel)
```
MichaelClerx, line 231: If you make the function return an emtpy list instead of `None`, these three lines can just be:
```
for param in self._get_list_of_parameters(sbml_model)
```
But it might be even nicer to have a method `self._parse_parameters(model)` ?

myokit/formats/sbml/_importer.py
MichaelClerx: Also, it seems you're building up some state in this parsing method, and storing it in the Parser instance. That's fine, but to make this re-usable I would suggest to either:

1. Split the parsing into several functions, that are each called with all the info they need. So any function that needs the userUnitDict would have it as an argument. This is my preferred option.

2. Alternatively, wrap all calls to `_parse_model` in a try/finally block that resets the instance back to its original state. So in `def parse()` you would have
```
try:
    self._parse_model(..)
...
finally:
    del self.userUnitDict
    del self.whatever
```

I'd try and do 1, and then where you really really can't avoid it use 2 as well

myokit/formats/sbml/_importer.py
```diff
@@ -35,334 +36,1065 @@ def __init__(self):
...
+                # Add size parameter to compartment
+                var = compDict[idx].add_variable('size')
+                var.set_unit(unit)
+                var.set_rhs(size)
+
+                # save size in container for later assignments/reactions
+                self.paramAndSpeciesDict[idx] = var
+
+        name = self._convert_name('myokit')
+        if 'MyoKit' in compDict:
```
MichaelClerx, line 224: I just spell it "Myokit" :D

myokit/formats/mathml/_parser.py
```diff
@@ -947,3 +951,13 @@ def _parse_name(self, element):
         except Exception as e:
             raise MathMLError('Unable to create Name: ' + str(e), element)
 
+    def _parse_symbol(self, element):
```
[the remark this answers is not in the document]
DavAug: The `csymbol` was previously not supported by the MathML parser right? Your first suggestion would be really cool in general, but maybe the latter one, just updating the docstring, would make more sense in this PR? So if you ever needed full `csymbol` support one could think about how to generalise this better. 
Or in other words, do you think you would want to add `csymbol` for other importers any time soon, and it would therefore make sense to worry about this now?

myokit/tests/data/formats/sbml/00004-sbml-l3v2-modified.xml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
DavAug: Haha ups. Will do!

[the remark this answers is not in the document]
DavAug: I added a LICENSE file, but the SBML file itself didn't have a license notice. The test suite GitHub repo, however, does have a [license](https://github.com/sbmlteam/sbml-test-suite/blob/master/LICENSE.html). Could you check out whether that is fine?

myokit/tests/test_formats_sbml.py
```diff
@@ -31,188 +31,770 @@
...
+    Tests the SBML importer. First correct construction of myokit model
+    is tested using an altered model from the SBML test suite. Then
+    handling of errors or misspecified files is checked.
     """
-
     @classmethod
     def setUpClass(cls):
         """
-        Test Hodgkin Huxley model.
+        Tests case 00004 from the SBML test suite
```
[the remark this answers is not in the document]
DavAug: Yes

myokit/tests/test_formats_sbml.py
```diff
@@ -31,188 +31,770 @@
...
         """
-        i = formats.importer('sbml')
-        cls.hohu = i.model(os.path.join(
-            DIR_FORMATS, 'sbml', 'HodgkinHuxley.xml'))
+        cls.i = formats.importer('sbml')
+        cls.modelFour = cls.i.parse_file(os.path.join(
+            DIR_FORMATS, 'sbml', '00004-sbml-l3v2-modified.xml'))
 
     def test_capability_reporting(self):
         """ Test if the right capabilities are reported. """
```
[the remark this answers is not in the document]
DavAug: haha no worries

myokit/tests/data/formats/sbml/00004-sbml-l3v2.mmt
```diff
@@ -0,0 +1,24 @@
+[[model]]
```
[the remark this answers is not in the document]
DavAug: Yep agreed. I used this to check the myokit model manually in the beginning and must have forgotten to delete it.

myokit/formats/sbml/_importer.py
DavAug: I have no idea :D the units in SBML lvl3 v2 are defined here

https://livermetabolism.com/paper/sbml-level-3-version-2-core.pdf

on page 44.

DavAug: " “item” is used for expressing such things as “N items” when “mole” is not an appropriate unit"

DavAug: So it seems to be quite generic. Is there an equivalent in myokit?

myokit/formats/sbml/_importer.py
DavAug: hmm I don't remember why I put it there, and it's not listed in the above base unit reference, so It may just be a mistake. I'll remove it.

myokit/formats/sbml/_importer.py
DavAug: Oh I see! This be may actually be useful for the web app, but maybe better to do in another PR :D

myokit/formats/sbml/_importer.py
DavAug: Oki doki. I played around with libsmbl too much so now my naming is all over the place. :D

myokit/formats/mathml/_parser.py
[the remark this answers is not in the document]
MichaelClerx: OK go for the docstring for now and I'll make a separate issue

myokit/formats/sbml/_importer.py
MichaelClerx: Nah, I'd just map it to dimensionless

myokit/formats/sbml/_importer.py
DavAug: Yes I refactored all logical blocks into smaller parsing methods.

myokit/formats/sbml/_importer.py
DavAug: Ah cool

DavAug:
Hi @MichaelClerx 
I refactored into an importer and a parser, but wasn't sure which interface with the parser would be preferable. So for now I just made it accessible through the importer interface which will have to be changed (this is what fails the tests).

myokit/formats/mathml/_parser.py
```diff
@@ -100,6 +100,9 @@ class MathMLParser(object):
         Becomes a :class:`myokit.Number`. To process units which may be present
         in the tag's attributes (esp. in CellML) the number post-processing
         function can be used.
+    ``<csymbol>``
```
MichaelClerx, line 103: Can you explain in this bit that the name generator is used? (The docstring below won't be visible to user)

myokit/formats/sbml/__init__.py
```diff
@@ -7,12 +7,14 @@
 from __future__ import absolute_import, division
 from __future__ import print_function, unicode_literals
 
-from ._importer import SBMLImporter, SBMLError  # noqa
+from ._importer import SBMLImporter  # noqa
+from ._parser import SBMLError, SBMLParser  # noqa
 
 
 # Importers
 _importers = {
```
MichaelClerx, line 15: I'd write this so that:

1. The SBMLImporter implements the Importer interface, and is the only class listed here
2. The SBMLImporter uses the SBMLParser, which doesn't implement any specific interface

myokit/formats/sbml/_parser.py
```diff
@@ -0,0 +1,1206 @@
...
+class SBMLError(myokit.ImportError):
+    """
+    Thrown if an error occurs when importing SBML
+    """
+
+
+# SBML base units according to libSBML
+sbml_to_myokit_unit_dict = {
+    'ampere': myokit.units.A,
+    'avogadro': None,  # TODO: myokit equivalent not yet looked up
```
MichaelClerx, line 1171: Hi @DavAug !

My final (?) comment is whether we need to catch this somewhere else. I'm guessing "1 avogadro" means "quite a lot", so treating this as dimensionless would be an error?

In that case, probably best to remove avogadro from this table, and have an explicit check before the lookup is done (if name == 'avogadro':) that raises an SBMLError explaining that this unit isn't supported

An alternative would be to actually implement this unit, as e.g. myokit.parse_unit('mol^-1 (6.02214076e23)'), if that's what the SBML definition is?

DavAug: It seems a unit avogadro is equivalent to a mole, and it just used for quantities that are not strictly speaking substances:

'Consequently, we avoid using “ mole” loosely where substances may not be
involved, and instead use “Avogadro’s number of X”. '

So it probably can't hurt to add it to the list of myokit units

DavAug: Actually now that I think about it, it's probably sufficient to just map avogadro to mole then

myokit/formats/sbml/_parser.py
```diff
@@ -0,0 +1,1206 @@
...
+class SBMLError(myokit.ImportError):
+    """
+    Thrown if an error occurs when importing SBML
+    """
+
+
+# SBML base units according to libSBML
+sbml_to_myokit_unit_dict = {
+    'ampere': myokit.units.A,
+    'avogadro': myokit.units.mol,  # This is a simplification of SBML's def
```
MichaelClerx, line 1171: The spec says to use

```(6.02214179 · 10 23 ) · dimensionless```

as the value, on page 44

In myokit that would be e.g. `myokit.parse_unit('1 (6.02214179e23)')`

File: myokit/formats/sbml/_parser.py [hunk 28]
```diff
@@ -0,0 +1,1206 @@
...
+    'becquerel': myokit.units.Bq,
+    'candela': myokit.units.cd,
+    'celsius': myokit.units.C,
```
MichaelClerx, line 1174: This one's not right. Myokit doesn't support celsius (or fahrenheit). So you'd have to catch this somewhere else and raise an exception if someone tries to use celsius

myokit/formats/sbml/_parser.py
```diff
@@ -0,0 +1,1206 @@
...
+    'kilogram': myokit.units.kg,
+    'liter': myokit.units.L,
+    'litre': myokit.units.L,
+    'lumen': myokit.units.lm,
+    'lux': myokit.units.lux,
+    'meter': myokit.units.m,
+    'metre': myokit.units.m,
+    'mole': myokit.units.mol,
+    'newton': myokit.units.N,
+    'ohm': myokit.units.R,
```
MichaelClerx, line 1195: Maybe just write `myokit.units.ohm` here. The `R` thing was a mistake I'll have to tidy up soon :-|

myokit/formats/sbml/_parser.py
DavAug: Oki doki

myokit/formats/sbml/_parser.py
DavAug: ok I wrapped the `smbl_to_myokit_dict` in a function that throws an error when celsius is passed

myokit/formats/sbml/_parser.py
DavAug: done

MichaelClerx:
Looks great. Merging it in!

MichaelClerx:
I'll try to do a myokit release later this week so that you can get the new SBML stuff via pypi !

MichaelClerx:
OK, so I notice I can no longer open the HH model file, because it's level 2 instead of 3 :D

Would it be a terrible idea to just warn about _possible_ incompatibility? I'm not sure how much SBML has changed between versions

DavAug:
Cool that was more work than I anticipated 😂

No I think it’s fine to just alert to possible incompatibilities. We just need to rewrite the get functions such that they allow a variety of namespaces.


On 26 May 2020, at 11:50, Michael Clerx <notifications@github.com> wrote:

﻿

OK, so I notice I can no longer open the HH model file, because it's level 2 instead of 3 :D

Would it be a terrible idea to just warn about possible incompatibility? I'm not sure how much SBML has changed between versions

MichaelClerx:
http://sbml.org/Documents/FAQ#What_are_the_SBML_Levels.2FVersions.2FReleases.2FRevisions_about.3F

This page shows the changes. Seems we definitely won't support 1, but should probably still support 2 (with warnings)

PR #547, opened 2020-05-26
Some further SBML updates

MichaelClerx:
- [x] Allow level 2 and level 3 (but warn if not latest) #543


MichaelClerx:
Closes #543

Issue #550, opened 2020-05-31
Make codecov work without a token


MichaelClerx:
See the way it's set up in PINTS

MichaelClerx:
https://docs.codecov.io/docs/frequently-asked-questions

MichaelClerx:
https://github.com/codecov/codecov-action/issues/29#issuecomment-594958141

PR #551, opened 2020-06-02
Removed codecov token

MichaelClerx:
See #550


MichaelClerx:
Works fine! So closes #550

Issue #549, opened 2020-05-31
Further refactoring for SBMLParser


MichaelClerx:
Several comments on SBML code, can be handled in a single PR or multiple PRs

## Replace dicts with classes

- [x] Add classes for e.g. compartment, species, parameters. These can then be used to store their properties (instead of having those in another dict)
- [x] Special parameters, e.g. the volume for a compartment should also be stored in that compartment, not in the central dict
- [x] Stop storing parameters and species in the same object
- [x] Conversion factors for species and model should be stored inside SBMLModel and Species classes, and not go in the main parameter/species/etc. namespace
- [x] This class should probably have some kind of dict or set of used ids, to check id uniqueness, while other things can be stored in separate containers (e.g. you'd have an `add_parameter(idx)` that  checks that the id is not used yet in the global id set, then creates a variable, stores it in e.g. `model._parameters[id] = var`, and then returns the var.)

## Reaction parsing
- [x] Parsing of reactions is a single very long method. Can we break this up into subsections? E.g. a seperate method to parse a reactant, and to parse a product?

## Error messages

- [x] Many error messages don't contain enough info to pinpoint errors at the moment, e.g. "Invalid id" without saying which id, or in which type of element
- [x] Line numbers can be provided by switching from ``elementtree`` to ``from lxml import etree`` (lxml is already a dependency of Myokit)

## Documentation

- [ ] We should have some more info in the SBMLParser docstring about how the conversion works, e.g. what components or variables get created when species etc. are used.
- [x] This docstring could also do with some more notes about what is and is not supported

## Tests
- [ ] Tests for species and model conversion factors
- [ ] Tests for stoichiometries defined as variables
- [ ] Any further tests needed to get to 100% coverage.
- [ ] In general, it would be nicer to test features separately inside SBMLParserTest, rather than combined with lots of other things in SBMLModelTest. Then you can have e.g. `test_assignment_rule()` which tests normal usage of assignment rules as well as errors
- [ ] The tests for overlapping ids could also just be part of the tests for species, compartments etc. (hopefully this breaks up a very long method)

## Extent units
```
# A problem occurs when the extentUnit and the species unit do
# not agree. Since initial values can be assigned to the
# species with substanceUnits, we will choose the species
# unit (in amount) over the globally defined extentUnits. This
# is NOT according to SBML guidelines.
```
[...]

PR #574, opened 2020-07-07
SBML parser refactoring

MichaelClerx:
See #549

- [x] Parser now only does parsing: stores parsed information into object structure.
- [x] Testing of parser can happen in smaller units now: doesn't need full model conversion anymore, so can be broken up into test_compartment, test_species, etc.
- [x] Conversion of parsed structure to myokit.Model is handled inside the sbml.Model method `myokit_model()`.

The sbml-to-myokit conversion will happen pretty much as before, but maybe with some slight changes.

Proposed strategy:

- Names are sids unless they starts with an underscore, then we need some resolution strategy
- Components are 'parameters', 'species', 'stoichiometries', etc., rather than compartments?
- Parameters with rate equations are ODEs, parameters without rate equations have their value() if set, otherwise initial_value()
- Stoichiometries all get a variable, set same way as parameters
- Species all get a variable for their amount, with a value set either from value(), initial_value(), or from a reaction
- Compartment sizes all get a variable, set same way as parameters
- Species that are concentrations get a 2nd variable, dividing by the compartment size

Question: ~Should there be a method that's called at the end of parsing that sets value/initial_value for all species based on
reactions? And~ If species are set with rules, does that set concentration or amount?


MichaelClerx:
@DavAug I've spent some time rewriting the parsing code, adding a clearer object structure (overlaps a bit with libsbml, I know), and learning about SBML. It should be possible to add much more modular unit tests now, see test_parse_compartment() for an example. Still needs lots of tests, and the actual myokit model creation isn't done yet!
Can you have a look at the code (easiest to download, PR viewer is a bit rubbish for this one) and the proposed strategy above and (1) see what you think, (2) let me know if you want to work on this some more

DavAug:
> See #549
> 
> * [x]  Parser now only does parsing: stores parsed information into object structure.
> * [ ]  Testing of parser can happen in smaller units now: doesn't need full model conversion anymore, so can be broken up into test_compartment, test_species, etc.
> * [ ]  Conversion of parsed structure to myokit.Model is handled inside the sbml.Model method `myokit_model()`.
> 
> The sbml-to-myokit conversion will happen pretty much as before, but maybe with some slight changes.
> 
> Proposed strategy:
> 
> * Names are sids unless they starts with an underscore, then we need some resolution strategy
> * Components are 'parameters', 'species', 'stoichiometries', etc., rather than compartments?
> * Parameters with rate equations are ODEs, parameters without rate equations have their value() if set, otherwise initial_value()
> * Stoichiometries all get a variable, set same way as parameters
> * Species all get a variable for their amount, with a value set either from value(), initial_value(), or from a reaction
> * Compartment sizes all get a variable, set same way as parameters
> * Species that are concentrations get a 2nd variable, dividing by the compartment size
> 
> Question: Should there be a method that's called at the end of parsing that sets value/initial_value for all species based on
> reactions? And if species are set with rules, does that set concentration or amount?

@MichaelClerx 
I just had a proper look at the refactored parser and it looks so much tidier than my version of it! Thanks a lot for all the work.

So if I understand it right, the parser is complete and just needs some testing, and I need to compose the myokit model from the already extracted information? 

I think your suggested strategy sounds good. I initially wanted to keep the compartment structure, because I think it would make visualizing the model structure easier (for the WebApp) especially for large models. But the Model object has all this information now, so we can minimize the number of components in the myokit.model.

For your questions, let me check the SBML documentation again.

MichaelClerx:
> So if I understand it right, the parser is complete and just needs some testing, and I need to compose the myokit model from the already extracted information?

That's exactly right!

I think the parser code is complete, and the testing is almost all there, except for assignment and rate rules, duplicate `SId`s, and anything the coverage testing throws up.

> I think your suggested strategy sounds good. I initially wanted to keep the compartment structure, because I think it would make visualizing the model structure easier (for the WebApp) especially for large models. But the Model object has all this information now, so we can minimize the number of components in the myokit.model.

I'm happy if you want to have a component per compartment too. That will probably be sensible for pkpd and general sys bio models, I suspect?
In that case I'd just go with any naming structure you think is sensible, and use `add_variable_allow_renaming` to "solve" any clashes (ugly if clashes occur, but as long as that's unlikely this is fine?), and maybe add an `add_component_allow_renaming` too?
The one thing I want to avoid is having to reject valid SBML files because they have SIds that overlap with some reserved name. So we should (a) avoid reserving names as much as possible; and (b) rename if clashes occur

> For your questions, let me check the SBML documentation again.
Thanks!
I'm glad you like the changes. I'm looking forward to seeing if the myokit model conversion code gets simpler now, or if this was all just tidying for the sake of tidying :D

DavAug:
>Question: Should there be a method that's called at the end of parsing that sets value/initial_value for all species based on
reactions?

I'm not quite sure I understand what you mean. If we are talking about the initial value of a species then according to the documentation is the initial value specified by either the initialAmount to initialConcentration attribute of the species, but can be overwritten with an initialAssignment. I don't think a Reaction can define the initial value of a species.

P. 54 SBML 3.1 "The value calculated by an InitialAssignment object overrides the value assigned to the given symbol by the object defining that symbol."

MichaelClerx:
> > Question: Should there be a method that's called at the end of parsing that sets value/initial_value for all species based on
> > reactions?
> 
> I'm not quite sure I understand what you mean. If we are talking about the initial value of a species then according to the documentation is the initial value specified by either the initialAmount to initialConcentration attribute of the species, but can be overwritten with an initialAssignment. I don't think a Reaction can define the initial value of a species.
> 
> P. 54 SBML 3.1 "The value calculated by an InitialAssignment object overrides the value assigned to the given symbol by the object defining that symbol."

Thanks! I can't remember what I meant either. Probably just confused :D

DavAug:
>And if species are set with rules, does that set concentration or amount?

According to the documentation it sets either.

P. 59 "In the case of a species, an AssignmentRule sets the referenced species’ quantity (whether a concentration or amount) to the value determined by the formula in math. The unit associated with the value produced by the math formula should be equal to the unit associated with the species’ quantity."

So whatever the species represents is set by the rule. What the species represents is specified by the hasOnlySubstanceUnits attribute, "false" meaning concentration and "true" meaning amount.

MichaelClerx:
> > And if species are set with rules, does that set concentration or amount?
> 
> According to the documentation it sets either.
> 
> P. 59 "In the case of a species, an AssignmentRule sets the referenced species’ quantity (whether a concentration or amount) to the value determined by the formula in math. The unit associated with the value produced by the math formula should be equal to the unit associated with the species’ quantity."
> 
> So whatever the species represents is set by the rule. What the species represents is specified by the hasOnlySubstanceUnits attribute, "false" meaning concentration and "true" meaning amount.

I'm wondering what happens if compartment sizes are also set.
I guess we could:

1. Make sure compartment initial assignments are handled before species initial assignments
2. Set species initial assignments for concentration to whatever **value** we calculate is appropriate
3. Set species assignments for concentrations to an **expression** `expression_given_by_rule / size_variable` ?

DavAug:
> > > And if species are set with rules, does that set concentration or amount?
> > 
> > 
> > According to the documentation it sets either.
> > P. 59 "In the case of a species, an AssignmentRule sets the referenced species’ quantity (whether a concentration or amount) to the value determined by the formula in math. The unit associated with the value produced by the math formula should be equal to the unit associated with the species’ quantity."
> > So whatever the species represents is set by the rule. What the species represents is specified by the hasOnlySubstanceUnits attribute, "false" meaning concentration and "true" meaning amount.
> 
> I'm wondering what happens if compartment sizes are also set.
> I guess we could:
> 
> 1. Make sure compartment initial assignments are handled before species initial assignments
> 2. Set species initial assignments for concentration to whatever **value** we calculate is appropriate
> 3. Set species assignments for concentrations to an **expression** `expression_given_by_rule / size_variable` ?

Sounds good to me.

Another thing:
Would it be ok, if I converted the ID according to your underscore_ + text rule already when we build up the model class?
In that way I don't need to worry about it anymore when building the myokit model, and the references in the Model class coincide with the references in the resulting myokit.model.

An alternative would be to to convert the IDs only when building the myokit.model, then there would be a mismatch between the Model and the myokit.model but the Model would reflect the SBML file definitions.

I'd prefer option one. Any thoughts?

MichaelClerx:
> > > > And if species are set with rules, does that set concentration or amount?
> > > 
> > > 
> > > According to the documentation it sets either.
> > > P. 59 "In the case of a species, an AssignmentRule sets the referenced species’ quantity (whether a concentration or amount) to the value determined by the formula in math. The unit associated with the value produced by the math formula should be equal to the unit associated with the species’ quantity."
> > > So whatever the species represents is set by the rule. What the species represents is specified by the hasOnlySubstanceUnits attribute, "false" meaning concentration and "true" meaning amount.
> > 
> > 
> > I'm wondering what happens if compartment sizes are also set.
> > I guess we could:
> > 
> > 1. Make sure compartment initial assignments are handled before species initial assignments
> > 2. Set species initial assignments for concentration to whatever **value** we calculate is appropriate
> > 3. Set species assignments for concentrations to an **expression** `expression_given_by_rule / size_variable` ?
> 
> Sounds good to me.
> 
> Another thing:
> Would it be ok, if I converted the ID according to your underscore_ + text rule already when we build up the model class?
> In that way I don't need to worry about it anymore when building the myokit model, and the references in the Model class coincide with the references in the resulting myokit.model.
> 
> An alternative would be to to convert the IDs only when building the myokit.model, then there would be a mismatch between the Model and the myokit.model but the Model would reflect the SBML file definitions.
> 
> I'd prefer option one. Any thoughts?

I was thinking it'd be nice if the Model had the original SIds, so that you could use them in error messages etc. without the user having to know about the name conversion. But I'm happy with whatever you decide

MichaelClerx:
For example the `_tag(self, element)` method shows IDs, as do the `__str__` methods of the Model/Compartment/etc. classes

DavAug:
Ok that makes sense! I guess the user can then explore the converted names in the myokit model herself/himself.

DavAug:
When parsing the initialAmount/initialConcentration I believe it's possible to specify the initial amount of a species indirectly by specifying the initialConcentration and setting the hasOnlySubstanceUnits to True. We would then have to convert the initial value using the compartment size. 

I'm wondering how to best do this, as the compartment size might be set by an initial assignment which we don't know about at the moment of extracting the species information.

MichaelClerx:
>  initialConcentration and setting the hasOnlySubstanceUnits

I thought that was an error?

DavAug:
> > initialConcentration and setting the hasOnlySubstanceUnits
> 
> I thought that was an error?

It certainly would make life easier if it were :D

I just checked the documentation again:

"The hasOnlySubstanceUnits attribute—Independently from how the initial value of a species’ quantity is set (Section 4.6.4), SBML allows choosing the meaning intended for a species’ identifier when the identifier appears in mathematical expressions or as the subject of SBML rules or assignments. The interpretation is controlled by the attribute hasOnlySubstanceUnits. If the attribute has the value “ false“, then the unit of measurement associated with the value of the species is {unit of amount}/{unit of size} (i.e., concentration or density). If hasOnlySubstanceUnits has the value “ true“, then the value is interpreted as having a unit of amount only."

DavAug:
"Although it may seem as though this intention could be determined by either (i) determining
whether whether the initialAmount or initialConcentration attribute is set on a
given Species object or (ii) examining the particular unit declared by the Species or Model
object’s substanceUnits attributes, the fact that all of these attributes are optional means
that a separate flag is needed. (Consider the situation where neither is set, and instead the
species’ quantity is established by an InitialAssignment or AssignmentRule—something
else is then needed to indicate whether the species’ identifier represents a concentration or
an amount.)"

DavAug:
I just realize, it's not super clear whether this excludes the possibility that the amount is indirectly specified through the initialConcentration. But the test example from the SBML test suite is doing exactly that, so I thought that we should just look at hasOnlySubstanceUnits?

MichaelClerx:
> I'm tempted to slightly adjust the Species class to just store both, the initial amount and the initial concentration, so we convert the initial value when creating the myokit.model.

OK, I think this is probably what we want then :-(

DavAug:
> > I'm tempted to slightly adjust the Species class to just store both, the initial amount and the initial concentration, so we convert the initial value when creating the myokit.model.
> 
> OK, I think this is probably what we want then :-(

Would it be ok to move the model species dict to the compartments? I would like to add the size of the compartment as a variable `size` and confirm in advance whether that variable name is already taken in the compartments name space.

MichaelClerx:
> Would it be ok to move the model species dict to the compartments? I would like to add the size of the compartment as a variable `size` and confirm in advance whether that variable name is already taken in the compartments name space.
> 
> For example in the HoHu model the size of the compartment would otherwise be called unit_component.unit_component which I find quite unintuitive

You could still call it "size"?

The species dict lives in the model because species names are unique model-wide, not component wide, that was my thinking!

DavAug:
> > Would it be ok to move the model species dict to the compartments? I would like to add the size of the compartment as a variable `size` and confirm in advance whether that variable name is already taken in the compartments name space.
> > For example in the HoHu model the size of the compartment would otherwise be called unit_component.unit_component which I find quite unintuitive
> 
> You could still call it "size"?
> 
> The species dict lives in the model because species names are unique model-wide, not component wide, that was my thinking!

That is true. I guess I am still trying to figure out the best way to find the myokit.model names when they deviate from the original references. For example species will have to come as an amount and potentially a concentration. But I'll just continue to play around with it and try to avoid unnecessary book-keeping.

MichaelClerx:
In the CellML-to-myokit conversion I just made a dict mapping cellml objects to myokit objects. Not pretty, but it only lived inside the `myokit_model()` method, so OK.
You could have a look at e.g. cellml.v2.api and then the `myokit_model()` method if that helps!

MichaelClerx:
@DavAug , for the species concentration / amount thing. Why don't we just store whatever the user specifies in the Species class, and only handle the split into substance and concentration when we generate the myokit model. Would that be a simpler option?

DavAug:
I can revert the changes though if you'd prefer to just let the parser do the parsing without interpreting the species initial amount

MichaelClerx:
I'll leave it up to you, but my idea was more to do things in the SBML way in the SBML classes, which should be quite natural? And then to let the convert-to-myokit method convert everything to something more myokit-y

DavAug:
> I'll leave it up to you, but my idea was more to do things in the SBML way in the SBML classes, which should be quite natural? And then to let the convert-to-myokit method convert everything to something more myokit-y
> 
> That way you can write the parsing etc. without having to worry about Myokit, and you won't have to give weird errors like "can't parse this model because it will lead to a weird thing when we convert it". That error would more naturally become something like "Error when converting: blah blah".
> 
> Then again, the only reason we're parsing in the first place is to convert to Myokit :D

Fair enough. Let me revert the changes then, but remove the import asserts that amount does not accept initialConcentration and the other way around. And then we do the conversion in the myokit_model() method. That is probably what you prefer no?

MichaelClerx:
Yeah that sounds good!

DavAug:
@MichaelClerx 
I've got one last problem and then the rest should go rather smoothly.

The parser constructs myokit expressions for the initialAssignment, assignmentRule, rateRule and kineticLaw by referring to the variables sid. That becomes problematic when the species is defined in concentration.

initialAssignment, assignmentRule and rateRule refer to the species in concentration (take species as it is), but the kineticLaw always refers to the amount. The myokit expressions do not distinguish between the two though. 

Is there an easy way to replace variable references in expressions? Say if the species as an sid and I call the amount myokit variable <sid>_amount, can I substitute the amount variable for sid easily?

DavAug:
This also makes renaming of the variables problematic, say calling the size of the compartment `size`.

MichaelClerx:
> Is there an easy way to replace variable references in expressions? Say if the species has an sid and I call the amount myokit variable 'sid'_amount, can I substitute the amount variable for sid easily?

Yeah you can do `Expression.clone(subst={old_expr: new_expr})`, for example `rhs = rhs.clone(subst={myokit.Name(x): myokit.Name(y)})`.
Does that help?

DavAug:
> > Is there an easy way to replace variable references in expressions? Say if the species has an sid and I call the amount myokit variable 'sid'_amount, can I substitute the amount variable for sid easily?
> 
> Yeah you can do `Expression.clone(subst={old_expr: new_expr})`, for example `rhs = rhs.clone(subst={myokit.Name(x): myokit.Name(y)})`.
> Does that help?

Yes thanks! I just saw that too in the documentation :d

DavAug:
@MichaelClerx 
Does the parsing of assignmentRules or rateRules handle explicit time dependence btw?

MichaelClerx:
> @MichaelClerx
> Does the parsing of assignmentRules or rateRules handle explicit time dependence btw?

Not something I've tested for! It will probably raise a KeyError when the "time" symbol is requested from a variable factory

DavAug:
> > @MichaelClerx
> > Does the parsing of assignmentRules or rateRules handle explicit time dependence btw?
> 
> Not something I've tested for! It will probably raise a KeyError when the "time" symbol is requested from a variable factory

I guess we could just leave it for now and worry about it in another release? But it's probably good to mention that time, delay and all these other csymbols are not yet supported.

MichaelClerx:
Wondering if we should stick all the tests for myokit_model in a separate class?

DavAug:
> Wondering if we should stick all the tests for myokit_model in a separate class?

I was wondering the same. I don't mind putting it in another class. I just thought that I'd be nice when the test class mirrors the true class. But there will be a lot of myokit_model tests, so your call

MichaelClerx:
Yeah best split it up then!

DavAug:
@MichaelClerx 
Just checking: in the 3.1. documentation it says the boundary attribute is called "boundaryCondition" and not "boundary". 

You used "boundary". is that a typo or has that changed from 3.1. to 3.2.?

DavAug:
@MichaelClerx 

Is it possible to use rates in myokit.Expression? SBML allows to use the sid of reactions to be referenced in mathematical expressions. Can we handle that?

DavAug:
@MichaelClerx 
I'm still missing a couple lines of coverage but you could start to have a look.

There is an example of how I thought we could include a couple examples from the SBML Test Suite, but the simulation results are marginally different. Potentially because they used a different ODE solver. I'm just wondering which degree of similarity is sufficient? 

As an example:
```
Mismatched elements: 43 / 51 (84.3%)
Max absolute difference: 8.46827181e-07
Max relative difference: 0.03068511
 x: array([0.0000000e+00, 1.3836376e-05, 2.6578291e-05, 3.8254924e-05,
       4.8895911e-05, 5.8562568e-05, 6.7411318e-05, 7.5467266e-05,
       8.2672521e-05, 8.9270838e-05, 9.5257967e-05, 1.0057254e-04,...
 y: array([0.0000000e+00, 1.4274387e-05, 2.7190387e-05, 3.8877267e-05,
       4.9451993e-05, 5.9020401e-05, 6.7678255e-05, 7.5512204e-05,
       8.2600655e-05, 8.9014551e-05, 9.4818084e-05, 1.0006934e-04,...
```

MichaelClerx:
Great stuff!
There seems to be a settings file containing info about what absolute and relative errors are acceptable for each case: https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/README.md#settings-files-for-timecourse-tests

Errors look a bit bigger in the above. Are we sure we're logging at the same time points?

DavAug:
> Great stuff!
> There seems to be a settings file containing info about what absolute and relative errors are acceptable for each case: https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/README.md#settings-files-for-timecourse-tests
> 
> Errors look a bit bigger in the above. Are we sure we're logging at the same time points?

Yes we are logging the same times. I included two more examples. One which agrees and another one which fails again. For the third example (case 01103), I would however say that the myokit result looks less like numerical errors than the provided solution.

```
Mismatched elements: 99 / 101 (98%)
Max absolute difference: 0.5
Max relative difference: 0.04761905
 x: array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9,  1. ,
        1.1,  1.2,  1.3,  1.4,  1.5,  1.6,  1.7,  1.8,  1.9,  2. ,  2.1,
        2.2,  2.3,  2.4,  2.5,  2.6,  2.7,  2.8,  2.9,  3. ,  3.1,  3.2,...
 y: array([ 0.    ,  0.1001,  0.2002,  0.3004,  0.4008,  0.5012,  0.6018,
        0.7025,  0.8032,  0.9041,  1.005 ,  1.106 ,  1.2072,  1.3084,
        1.4098,  1.5112,  1.6128,  1.7145,  1.8162,  1.9181,  2.02  ,...
```

`x` is the solution from myokit.

MichaelClerx:
Trying to read that model, does it say:
- X is created with rate k1=0.01 and stoichiometry Xref
- Xref(t0) = 1, dot(Xref)=0.01
?

DavAug:
> Trying to read that model, does it say:
> 
> * X is created with rate k1=0.01 and stoichiometry Xref
> * Xref(t0) = 1, dot(Xref)=0.01
>   ?

I believe so yes. I wanted to check whether some more complicated reactions work as well

MichaelClerx:
Yeah my first guess is we aren't updating the stoichiometry?

DavAug:
> Yeah my first guess is we aren't updating the stoichiometry?

The myokit model looks ok though no?

```
ame: case01103
# Initial values
default_compartment.Xref     = 1
default_compartment.X_amount = 0

[default_compartment]
dot(X_amount) = myokit.k1
    in [1]
X_concentration = X_amount / size
    in [1]
dot(Xref) = 0.01
    in [1]
size = 1
    in [1]

[myokit]
k1 = 1
time = 0 bind time
    in [1]
```

MichaelClerx:
>>> import numpy as np
>>> x = np.arange(0, 10, 101)
>>> x
array([0])
>>> x = np.linspace(0, 10, 101)
>>> x
array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9,  1. ,
        1.1,  1.2,  1.3,  1.4,  1.5,  1.6,  1.7,  1.8,  1.9,  2. ,  2.1,
        2.2,  2.3,  2.4,  2.5,  2.6,  2.7,  2.8,  2.9,  3. ,  3.1,  3.2,
        3.3,  3.4,  3.5,  3.6,  3.7,  3.8,  3.9,  4. ,  4.1,  4.2,  4.3,
        4.4,  4.5,  4.6,  4.7,  4.8,  4.9,  5. ,  5.1,  5.2,  5.3,  5.4,
        5.5,  5.6,  5.7,  5.8,  5.9,  6. ,  6.1,  6.2,  6.3,  6.4,  6.5,
        6.6,  6.7,  6.8,  6.9,  7. ,  7.1,  7.2,  7.3,  7.4,  7.5,  7.6,
        7.7,  7.8,  7.9,  8. ,  8.1,  8.2,  8.3,  8.4,  8.5,  8.6,  8.7,
        8.8,  8.9,  9. ,  9.1,  9.2,  9.3,  9.4,  9.5,  9.6,  9.7,  9.8,
        9.9, 10. ])
>>> y = x + 0.005*x**2

MichaelClerx:
Good find!

Do you think it would make sense to put the model classes etc. in a separate `_api.py` module like we have for CellML? It'd mean Model etc. all become public (e.g. `myokit.formats.sbml.Model` etc.), so we'd need to add them to the docs table of contents.
The tests could then get split over e.g. `myokit/tests/test_formats_sbml.py` (importer), `myokit/tests/test_sbml_parser.py`, `myokit/tests/test_sbml_api.py` (which would just test the model conversion code, I guess).

Or is that too much?

I guess it should also be fairly straightforward to do a `Model.from_myokit_model()` at some point (no reactions, just creates parameters with (rate) equations), and an exporter...
But not relevant for your PKPD project I guess :D

MichaelClerx:
Yeah let's not think about export just yet. Not until there's somebody who needs it anyway :D

It's looking good @DavAug !

DavAug:
> Good find!
> 
> Do you think it would make sense to put the model classes etc. in a separate `_api.py` module like we have for CellML? It'd mean Model etc. all become public (e.g. `myokit.formats.sbml.Model` etc.), so we'd need to add them to the docs table of contents.
> The tests could then get split over e.g. `myokit/tests/test_formats_sbml.py` (importer), `myokit/tests/test_sbml_parser.py`, `myokit/tests/test_sbml_api.py` (which would just test the model conversion code, I guess).
> 
> Or is that too much?
> 
> I guess it should also be fairly straightforward to do a `Model.from_myokit_model()` at some point (no reactions, just creates parameters with (rate) equations), and an exporter...
> But not relevant for your PKPD project I guess :D

I think it would be really cool to be able to convert back to SBML, but for the PKPDApp it would be more useful to also know about reactions etc. So I guess it would make more sense to reconstruct the SBML file from the Model class in this case, and define an interface to update the Model class...

Anyway let's focus on the Importer / Parser here.

I think it would be super helpful if the Model class would be public! :)

DavAug:
So would that mean that we just put the Model class in a separate `_api.py` file and add it to the doc?

MichaelClerx:
Yup! And import all the classes (and methods) you need from the _api inside `myokit.formats.sbml.__init__`, and then make sure that the `_parser` module accesses them via `myokit.formats.sbml`.

DavAug:
@MichaelClerx 
There seem to be some issues with python 2.7 when simulating the test cases. Could you have a look at that? I can't really see what the problem is.

MichaelClerx:
In Python 2 it generates a model like this:
```
[[model]]
name: case00004
# Initial values
compartment.S2_amount = 0
compartment.S1_amount = 0.15

[compartment]
dot(S1_amount) = -(1 * (size * myokit.k1 * S1_concentration))
    in [mol]
S1_concentration = S1_amount / size
    in [M]
dot(S2_amount) = 2 * (size * myokit.k1 * S1_concentration)
    in [mol]
S2_concentration = S2_amount / size
    in [M]
size = 1
    in [L]

[myokit]
k1 = 0.35
k2 = 180
time = 0 bind time
    in [s]
```
instead of this
```
[[model]]
name: case00004
# Initial values
compartment.S1_amount = 0.15
compartment.S2_amount = 0

[compartment]
dot(S1_amount) = -(1 * (size * myokit.k1 * S1_concentration)) + 1 * (size * myokit.k2 * S2_concentration ^ 2)
    in [mol]
S1_concentration = S1_amount / size
    in [M]
dot(S2_amount) = 2 * (size * myokit.k1 * S1_concentration) - 2 * (size * myokit.k2 * S2_concentration ^ 2)
    in [mol]
S2_concentration = S2_amount / size
    in [M]
size = 1
    in [L]

[myokit]
k1 = 0.35
k2 = 180
time = 0 bind time
    in [s]
```

MichaelClerx:
I think we should rename the methods like `x()` **which return a boolean** to `is_x()`, e.g. `rate() --> is_rate()`.

MichaelClerx:
Something, somewhere, depended on the ordering of the compartments/reactions/parameters...
Fixed that by using OrderedDict which is ordered in both 2 and 3.
But seems a bit suspicious that part of the code requires this ordering?

MichaelClerx:
So current todo is

- [ ] figure out why the myokit_model code needed
- [ ] rename the methods like x() which return a boolean to is_x(), e.g. rate() --> is_rate().
- [ ] split api off into a separate module, expose via `myokit.formats.sbml`
- [ ] split the tests for the api off into their own test module
- [ ] get coverage up to 100%

You happy to tackle those @DavAug ?

DavAug:
@MichaelClerx it probably makes sense to write the api tests independent of the parser doesn't it? I.e. setting spatial dimensions and then checking that those dimensions are correct?

MichaelClerx:
How's it going @DavAug ? I see you've added a lot of new tests, but also that travis doesn't pass so that coverage testing isn't working? Might be good to fix that first so that you can isolate crucial points easier?
Happy to help if you need anything!

DavAug:
> How's it going @DavAug ? I see you've added a lot of new tests, but also that travis doesn't pass so that coverage testing isn't working? Might be good to fix that first so that you can isolate crucial points easier?
> Happy to help if you need anything!

Oh I didn't realise that Travis stopped working. :D

Apart from that it's going ok though. I just felt that testing the api through the parser is not ideal. That is why I started testing the api separate from the parser.

MichaelClerx:
Cool! Looks like there's still a lot of parsing happening in the API test though? Would it not be easier to build the model programmatically in those tests to avoid testing the parser at the same time?

DavAug:
> Cool! Looks like there's still a lot of parsing happening in the API test though? Would it not be easier to build the model programmatically in those tests to avoid testing the parser at the same time?

Agreed! I haven't touched this test yet. I will change it though.

DavAug:
Apparently it doesn't work to import sbml with 

```
import myokit.formats.sbml 
```

It works on my mac though. Do you know why that is?

MichaelClerx:
> Apparently it doesn't work to import sbml with
> 
> ```
> import myokit.formats.sbml 
> ```
> 
> It works on my mac though. Do you know why that is?

Not sure! I'd maybe start by deleting any __pychache__ directories, and seeing if it still works

DavAug:
> > Apparently it doesn't work to import sbml with
> > ```
> > import myokit.formats.sbml 
> > ```
> > 
> > 
> > It works on my mac though. Do you know why that is?
> 
> Not sure! I'd maybe start by deleting any **pychache** directories, and seeing if it still works

Hmm, I have gitignored my pychaches, so that wouldn't help with Travis would it?

MichaelClerx:
> Hmm, I have gitignored my pychaches, so that wouldn't help with Travis would it?

Sometimes you're importing an older, deleted file's cache! But that wasn't happening in this case

DavAug:
> > Hmm, I have gitignored my pychaches, so that wouldn't help with Travis would it?
> 
> Sometimes you're importing an older, deleted file's cache! But that wasn't happening in this case

It's weird because in commit 884995d some python versions pass and others don't.

Do you know whether imports somehow changed between the versions? I'm quite lost :D the __init__.py are there so it should work.

MichaelClerx:
I think I fixed it!

DavAug:
> I think I fixed it!

haha but why couldn't I import the SBMLError? :D

MichaelClerx:
Not sure! I moved the error mostly so that the API didn't depend on the parser.
The issue in the parser seemed to be that `import x as y` is apparently different from `import x`. Maybe that was fixed in 3.7+ ?

MichaelClerx:
Also needs a bit of thinking about SBMLError vs SBMLParserError, as in, should the parser only raise parser errors? I think that was the design, no? Can't quite remember! Whatever the resolution, it should probably be consistent with the use of these errors in the cellml api and parser code :-)

DavAug:
> Also needs a bit of thinking about SBMLError vs SBMLParserError, as in, should the parser only raise parser errors? I think that was the design, no? Can't quite remember! Whatever the resolution, it should probably be consistent with the use of these errors in the cellml api and parser code :-)

That is how I interpreted the SBMLParserError. Are there still places in the parser where an SBMLError is thrown?

DavAug:
> > Also needs a bit of thinking about SBMLError vs SBMLParserError, as in, should the parser only raise parser errors? I think that was the design, no? Can't quite remember! Whatever the resolution, it should probably be consistent with the use of these errors in the cellml api and parser code :-)
> 
> That is how I interpreted the SBMLParserError. Are there still places in the parser where an SBMLError is thrown?

Ok yes only in the parse_mathml function the SBMLError is still thrown. So do we want an SBMLParserError instead?

MichaelClerx:
> > Also needs a bit of thinking about SBMLError vs SBMLParserError, as in, should the parser only raise parser errors? I think that was the design, no? Can't quite remember! Whatever the resolution, it should probably be consistent with the use of these errors in the cellml api and parser code :-)
> 
> That is how I interpreted the SBMLParserError. Are there still places in the parser where an SBMLError is thrown?

Nope! Sorry! Just confusing myself

DavAug:
@MichaelClerx 

The sourceline attribute does not exist for xml.Elementtree.Element, only in lxml. So how important is this for you to work

```
class SBMLParsingError(myokit.ImportError):
    """
    Thrown if an error occurs when parsing SBML.

    The argument ``element`` can be used to pass in an element that caused the
    error.
    """
    def __init__(self, message, element=None):
        if element is not None:
            try:
                line = str(element.sourceline)
                message = 'Error on line ' + line + '. ' + message
            except AttributeError:  # pragma: no cover
                pass
        super(SBMLParsingError, self).__init__(message)
```

Can we just replace xml by lxml?

MichaelClerx:
I've switched the parser code to use lxml, as this is a dependency for myokit now!

MichaelClerx:
For the tests, could you split them into
- `test_formats_sbml.py` - really basic tests of the importer, only checks extra functionality that importer adds
- `test_sbml_api.py` - API tests
- `test_sbml_parser.py` - Parser tests

This would make it consistent with the cellml tests

DavAug:
@MichaelClerx 
I'm trying to increase the coverage of the parser and I'm wondering whether the parser even allows xml files without a namespace? Also I think the tag _method is only used on elements that do not belong to the mathml_ns, at list I somehow struggle to come up with an example. I'm tempted to remove the ifs. i.e. 

```
def _tag(self, element):
        """
        Returns an element's name, but changes the syntax from ``{...}tag`` to
        ``sbml:tag`` for SBML elements.
        """
        ns, el = myokit.formats.xml.split(element.tag)

        # Replace known namespaces
        if ns is None:
            tag = el
        elif ns == self._ns:
            tag = 'sbml:' + el
        elif ns == NS_MATHML:
            tag = 'mathml:' + el
        else:
            tag = '{' + ns + '}' + el

        # Add id if known
        sid = element.get('id', element.get('unitsid'))
        if sid is not None:
            tag += '[@id=' + str(sid) + ']'

        return tag
```

becomes

```
def _tag(self, element):
        """
        Returns an element's name, but changes the syntax from ``{...}tag`` to
        ``sbml:tag`` for SBML elements.
        """
        ns, el = myokit.formats.xml.split(element.tag)

        # Replace known namespaces
        if ns == self._ns:
            tag = 'sbml:' + el
        else:
            tag = '{' + ns + '}' + el

        # Add id if known
        sid = element.get('id', element.get('unitsid'))
        if sid is not None:
            tag += '[@id=' + str(sid) + ']'

        return tag
```

DavAug:
@MichaelClerx 
On last question:
I got the coverage of the api up to 100%, but I am now rewrtitng the tests for the myokit_model method. I find it very difficult to write good unit tests for this method because it's so long ...

Do you think it would make sense to break it up into private methods that perform the various steps of the sbml to myokit translation, i.e. `_add_components`, `_set_size_values` etc. That way I could easier test the methods and maybe myokit_model also becomes more readable.

DavAug:
@MichaelClerx 
Ok one more question :D
Should the methods of the various api classes have a more detailed explanation of their arguments?

MichaelClerx:
> @MichaelClerx
> I'm trying to increase the coverage of the parser and I'm wondering whether the parser even allows xml files without a namespace? Also I think the tag _method is only used on elements that do not belong to the mathml_ns, at list I somehow struggle to come up with an example. I'm tempted to remove the ifs. i.e.
> 
> ```
> def _tag(self, element):
>         """
>         Returns an element's name, but changes the syntax from ``{...}tag`` to
>         ``sbml:tag`` for SBML elements.
>         """
>         ns, el = myokit.formats.xml.split(element.tag)
> 
>         # Replace known namespaces
>         if ns is None:
>             tag = el
>         elif ns == self._ns:
>             tag = 'sbml:' + el
>         elif ns == NS_MATHML:
>             tag = 'mathml:' + el
>         else:
>             tag = '{' + ns + '}' + el
> 
>         # Add id if known
>         sid = element.get('id', element.get('unitsid'))
>         if sid is not None:
>             tag += '[@id=' + str(sid) + ']'
> 
>         return tag
> ```
> 
> becomes
> 
> ```
> def _tag(self, element):
>         """
>         Returns an element's name, but changes the syntax from ``{...}tag`` to
>         ``sbml:tag`` for SBML elements.
>         """
>         ns, el = myokit.formats.xml.split(element.tag)
> 
>         # Replace known namespaces
>         if ns == self._ns:
>             tag = 'sbml:' + el
>         else:
>             tag = '{' + ns + '}' + el
> 
>         # Add id if known
>         sid = element.get('id', element.get('unitsid'))
>         if sid is not None:
>             tag += '[@id=' + str(sid) + ']'
> 
>         return tag
> ```

👍🏻 👍🏻 👍🏻

MichaelClerx:
> @MichaelClerx
> One last question:
> I got the coverage of the api up to 100%, but I am now rewrtitng the tests for the myokit_model method. I find it very difficult to write good unit tests for this method because it's so long ...
> 
> Do you think it would make sense to break it up into private methods that perform the various steps of the sbml to myokit translation, i.e. `_add_components`, `_set_size_values` etc. That way I could easier test the methods and maybe myokit_model also becomes more readable.
> 
> It would have the disadvantage that we would have to pass the dictionaries around, which potentially makes it slower for really big models

Up to you! I'm happy either way

MichaelClerx:
> @MichaelClerx
> Ok one more question :D
> Should the methods of the various api classes have a more detailed explanation of their arguments?

A nice list of arguments is never wrong!
But personally if it's a simple function I'm happy with something like ```Sets the cell size to ``size``.```, or ```Returns the sum of ``x`` and ``y``.```

DavAug:
@MichaelClerx 
With the last commit the coverage should be up to 100%, but I also removed all if statements from the _tag method. I believe the way the _path method is constructed we will only look at elements under the defined namespace `self._ns` (and the mathml namespace). At least I weren't able to construct examples that could trigger those if statements.

MichaelClerx:
I'm happy to merge now @DavAug ! Let me know when you are

DavAug:
> I'm happy to merge now @DavAug ! Let me know when you are

There are a couple tests that I didn't refactor yet. I'll finish them up tonight and then I'm happy too :)

DavAug:
@MichaelClerx the docs test complaints about referencing the myokit.formats.sbml module twice.

MichaelClerx:
@DavAug fixed it!

DavAug:
@MichaelClerx 
I am half way through with refactoring the tests for the myokit_model method, such that they do not rely on the parser anymore. Just let me know whether you think it's worth the effort, or whether you are happy to keep this dependence of the api tests.

MichaelClerx:
If you get fed up with it then stop at any time :D
We can always leave it as a separate issue

DavAug:
> If you get fed up with it then stop at any time :D
> We can always leave it as a separate issue

haha yes I'm a little bored right now

MichaelClerx:
@DavAug just looking at the API tests, it seems they test a lot of things that they don't need to?
For example, `test_stoichiometry_values` sets stoichiometries in different ways, but this is all covered by any test for Quantity, no?
Similarly a lot of the methods check the myokit_model() conversion output as well, which is probably a duplicate of the main tests for myokit_model() ?

MichaelClerx:
So might be faster to just remove a load of them?

DavAug:
> So might be faster to just remove a load of them?

Agreed, I essentially was planning to replace all tests for the myokit_model method that start with an sbml file and just test the subroutines `_add_components`, `_set_rhs_sizes` etc.

I can remove some of them, but not yet all otherwise there will be likely untested code.

MichaelClerx:
Hmmm! That suggests we should move some of the code to either parser testing or the myokit_model tests?

DavAug:
> Hmmm! That suggests we should move some of the code to either parser testing or the myokit_model tests?

Sorry, I meant that removing some of the myokit_model tests that still use the parser will likely lead to leave some of the myokit_model method untested, but that's just because I didn't finish refactoring the tests. Does that make sense?

MichaelClerx:
I've merged it in anyway!

New issue for the final bits #598

Issue #44, opened 2018-02-23
Add simulation with CVODES to calculate sensitivities.


MichaelClerx:

MichaelClerx:
Or with sympy export/import

MichaelClerx:
@chonlei is looking into this now. Might be possible without knowing df/dp ?

MichaelClerx:
From the CVODES docs (2.7):

> There are several methods for evaluating the right-hand side of the sensitivity systems (2.11): analytic evaluation, automatic differentiation, complex-step approximation, and finite differences (or directional derivatives).
> CVODES provides all the software hooks for implementing interfaces to automatic differentiation (AD) or complex-step approximation; future versions will include a generic interface to AD-generated functions.
> At the present time, besides the option for analytical sensitivity right-hand sides (user-provided), CVODES can evaluate these quantities using various finite difference-based approximations to evaluate the terms (∂f /∂y)s i and (∂f /∂p i ), or using directional derivatives to evaluate [(∂f /∂y)s i + (∂f /∂p i )].
> As is typical for finite differences, the proper choice of perturbations is a delicate matter.
CVODES takes into account several problem-related features: the relative ODE error tolerance rtol, the machine unit roundoff U , the scale factor p̄ i , and the weighted root-mean-square norm of the sensitivity vector s i .

PR #577, opened 2020-07-14
Sensitivity calculation with CVODES

MichaelClerx:
See #44 

- [x] Write a plain C example that uses FSA and finite differences to calculate sensitivities w.r.t. parameters
- [x] Update the example to also calculate sensitivities w.r.t. one of the initial conditions
- [ ] Create a new class, e.g. `SSimulation` or `SimulationS1` etc. Like PSimulation, it should have constructor arguments `variables`, and `parameters`, but also `initial_conditions`. run() should return a nice data log or a tuple.
- [ ] Work out if/how to support calculation of e.g. dI/dp|t
- [ ] Add tests
- [ ] Update docs and web example
- [ ] Deprecate PSimulation and ICSimulation (but what about users with CVODE but no CVODES?)
- [ ] Test on windows. Do I need to compile new CVODES dlls?


MichaelClerx:
Still need a way to get dI/dp, for arbitrary I and p...

This might be the way:
```
#!/usr/bin/env python3
import myokit

m = myokit.load_model('example')
i = m.get('ina.INa')

es, args = m.expressions_for(i)

def collapse(es):
    subst = {}
    for lhs, rhs in es[:-1]:
        if not rhs.is_literal():
            subst[lhs] = rhs.clone(subst=subst)
    return es[-1].rhs.clone(subst=subst)

single_expr = collapse(es)

import myokit.formats
w = myokit.formats.ewriter('sympy')
e = w.ex(single_expr)

print(e)

import sympy as sp
print(sp.diff(e, w.ex(myokit.Name(m.get('ina.m')))))
```
returns
```
3.0*ina.gNa*ina.h*ina.j*ina.m**2.0*(membrane.V - R*T*log(cell.Na_o/cell.Na_i)/F)
```
so could read that back in and use to calculate dI/dm, then multiply by dm/dp

PR #581, opened 2020-07-22
Naive partial_derivative method for expressions

MichaelClerx:
Addresses #583

## Main ideas:

- `PartialDerivative` is an `LhsExpression` that represents a partial derivative
- `InitialValue` is an `LhsExpression` that represents an initial value, but can only be used in a `PartialDerivative(Name, InitialValue)` (this is checked by `_validate()` methods, not at construction).
- Both `InitialValue` and `PartialDerivative` have a code representation (`init(x)` and `partial(x, y)`), but no parsing will be implemented (for now)
- `Expression` has a `partial_derivative(lhs)` method, that returns the derivative of the expression w.r.t. the `Name` or `InitialValue`
- The `derivative(lhs)` method performs very very slight simplification: Each expression has a `_partial_derivative(lhs)` method that can return None for zero, and all methods are None-aware. If a None reaches the top level (`partial_derivative()`) it is turned into a zero with the correct units
- `_partial_derivative()` methods add the correct units for any `Number`s they introduce
- The `_partial_derivative()` method of `myokit.Name` should return a `PartialDerivative` object if its variable depends on the requested `lhs`. To implement this, the `partial_derivative` method will temporarily add a binding for `lhs` (if it's constant), so that the variable's `is_constant()` method can be used.

More changes:
- Unit and Quantity have their own module `_unit.py` now
- Added methods `Expression.is_name(var=None)` and `Expression.is_number(value=None)`.
- Moved `is_derivative()` from `LhsExpression` to `Expression` and added argument `var=None`
- `Expression.contains_type(PartialDerivative)` and `Expression.contains_type(InitialValue)` will be used a lot (e.g. to check that a variable RHS does _not_ contain them) so get special caching
- Might have found some issues with uniqueness of names of user functions (which nobody uses)

Future plans:
- InitialValue and PartialDerivative can be used in output for CVODES simulation
- Might be useful at some point to also allow parsing of `partial()` and `init()` constructs.
- Maaaaaybe the PartialDerivative class can be used when importing CellML, so that PDE models can be represented - but not solved - in Myokit. The validate() method would still raise errors, as the RHS of a variable cannot contain a PartialDerivative
- The InitialValue class doesn't help CellML import, as CellML has no similar construct. For this we would need to allow `init(x) = p`, so that `d/dinit(x)` becomes `d/dp`. This is how the CVODES docs formulate it too. But still not certain this would be the way to go.


MichaelClerx:
Thinking about initial values a bit more. It might be quite nice to solve this by letting initial values depend on constants. Then instead of `partial(x, init(y))` we'd get `partial(x, p)` so no special syntax there beyond the `partial` keyword.
Have been pushing to get init = f(p) out of CellML though... (it's still in).
Validation doesn't seem too hard: just check that any variable referenced in an initial value is a constant.

Much bigger problem is the idea that a model state can be saved, loaded, etc. That doesn't sit well with the idea of the initial condition depending on the model. I do still like keeping the model as a mapping/orbit on the state space, and the initial values as a point

MichaelClerx:
@DavAug no need to read all of the above, just this to-do list : D

- [x] Implementation
- [x] Tests
- [ ] Full test coverage
- [ ] Add minimal simplification for 1 * , when dx/dx returns a 1 [1]
- [ ] List all big changes/additions in changelog

Issue #593, opened 2020-08-05
cvodessim: Change `variables` in sensitivity list to also accept `dot(x)`


MichaelClerx:
Allow logging of d/dp dot(x) (syntax is there in .py already)

MichaelClerx:
Models without derivatives can't have sensitivities: only allowed for state, deriv, or intermediary (= state dependent)

Issue #592, opened 2020-08-05
Move most code out of templates


MichaelClerx:
Things that should go out of the template

- model pre-manipulation
- solvable order
- declaring a name function
- making big lists of things

Things that should stay in
- printing equations
- easy for loops
- etc

Ideally, anything complicated (even calls to `w.eq(eq)` or `v(var)`) should go in the python code, where there's better error handling and test coverage checking

PR #595, opened 2020-08-07
CVODES Simulation to calculate sensitivities

MichaelClerx:
See #44. Replaces the main Simulation type, as well as PSimulation and ICSimulation

## Phase 1 - derivatives
- Implement derivatives in the expression system
- #581 Get full coverage and merge

## Phase 2 - usable prototype
- [x] Implement basic cvodes sim with finite differences
- [x] Replace cvodesim with cvodessim
- [x] Drop `set_constant` from c simulation, instead pass constant values to C in sim_init (and in eval derivs method)
- [x] Move any C code that uses the model into a separate C function
- [x] Make root finding something you can specify per run
- [x] Define a struct that holds all model vars, states, derivs(, jac, sens), use in methods (Logging is part of model code!)
- [x] Work out best way to store variables, sensitivities etc inside model struct. Should sensitivities be in an array too? Then easier to iterate over, can still define 
- [x] Tidy Python code (split long methods)
- [x] Make sure Python code does not reorder deps and indeps
- [x] Make sure C logging code maintains order of deps and indeps
- [x] Move model code into separate file - even if only for readability -->  New issue #698
- [x] Move most python code from template to .py file (part of #592)

## Phase 3 - proper version
- [x] Make sure sensitivities are stored on start/restart
- [x] Allow sim to stop and restarted, appending to a sensitivity matrix list
- [x] Figure out final CModel interface and calling order
- [x] Update cvodes code to use new CModel interface
- [x] Look for any changes in cvode sim since this branch started, incorporate them in new code
- [x] #703 Check state passing in and out of sim
- [x] Try running with caching

## Phase 3.5 - edge cases
- [ ] #593 Allow logging of d/dp dot(x) (syntax is there in .py already)
- [ ] Model without states. What happens for e.g. `x = sin(p * time)`, sens w.r.t `p` ?

## Phase 4 - testing and platforms
- [x] Add numerical test for simulation, based on analytical solution of system of ODES
- [x] Add numerical test for sensitivities, based on analytical solution of system of ODES
- [x] Pass start/stop and event logging tests (strengthen if needed)
- [x] Test pickling
- [ ] Add tests for full coverage
- [x] CVODES binaries for windows
- [x] Test (somehow) with CVODES 2,3,4,5
- [ ] Test sensitivity calculations on big models
- [x] Benchmark new Simulation class against old one (compilation is 10% slower, but not sure about run time)
- [x] Do we need readable code e.g. all the aliases? Maybe remove and benchmark compilation with & without?

## Phase 5: Rest of myokit
- [x] Search for "cvode" in all docs, code, licenses, update to cvodes where required
- [x] Remove or deprecate PSimulation and SimulationIC
- [x] Update changelog


MichaelClerx:
## Memoisation

All model variables live in a struct, including states, derivatives, sensitivities etc. None are in vectors

```
int rhs(t, y, user_data) {
  determine if t, y, user_data match static data
  if (needed) {
    unpack t, y, user_data onto static
    evaluate rhs equation
  }
  return needed;
}
int jac(t, y, user_data) {
  if (rhs(t, y, user_data)) {
    re-evaluate jac
  }
}
int sens(t, y, sy, user_data) {
  if (jac(t, y, user_data)) {
    re-evaluate sens
  }
}

cvode_rhs(t, y, ydot, user_data) {
  rhs(t, y, user_data);
  pack static dy into ydot vector;
}
cvode_jac(t, y, ydot, jac, user_data) {
  jac(t, y, user_data);
  pack static dy into ydot vector;
}
cvode_sens(t, y, ydot, yS, ySdot, user_data) {
  sens(t, y, yS, user_data);
  pack static ySdot into ySdot vector;
}
```

MichaelClerx:
Note to self: Some changes have been made in CVODE sim w.r.t. handling very short sim times. Need to be incorporated here.

MichaelClerx:
Now wondering if memoisation should be inside the model, not inside the simulator functions.

That way they get shared across classes using the same model code

But also, it doesn't make sense to call RHS functions with fake args instead of just calling the model methods directly

MichaelClerx:
Using a model:

Start:
```
model = Model_Create() -> Sets all bound and literal variables
```

After any change to literals: (e.g. before start, or before re-start)
```
Model_EvaluateLiteralDerivedVariables() --> Sets literal-derived variables
```

During simulation
```
Model_SetBoundVariables(time, pace) -> Sets bound, clears cache if they're changed
Model_SetParameters(*parameters) -> Sets parameters, clears cache if they've changed
Model_SetState(*state) -> Sets state, clears cache if it's changed

Model_EvaluateDerivatives() -> Does nothing if cache exists, otherwise re-evaluates derivatives of state
Model_EvaluateSensitivityOutputs() {} -> Does nothing if cache exists, otherwise re-evaluates derivatives of sensitivities dy/dx where y is an intermediary variable
```

myokit/_sim/cmodel.py
```diff
@@ -0,0 +1,428 @@
...
+            elif not var.is_literal():
+                raise ValueError(
+                    'Sensitivity with respect to ' + var.qname() +
+                    ' requested, but this is not a literal variable (it'
+                    ' depends on other variables).')
+            independents.append(lhs)
+
+        return True, dependents, independents
+
+    def _derive_sensitivity_equations(self, model, deps, indeps):
```
DavAug, line 242: This function seems to always return empty lists. Is it possible that in line 256 it was supposed to be 
```
output_variables = [
            lhs.var() for lhs in deps
            if (isinstance(lhs, myokit.Name) and lhs.var().is_state())]
```

DavAug: Ahh I just read the comment above, so you intentionally ignore states?

MichaelClerx: Yeah, the d/dp of the state variables is already known at this point, but in cardiac models we usually have a lot of variables of interest that aren't state variables, but some function of them instead. So this method is meant to find the equations to calculate the sensitivities of those "intermediate variables"

DavAug:
It may sound silly but would you mind adding one or two small tests to capture the on DataLog line and the the realtime associated block in the LegacySimulation.run @MichaelClerx ? I have a hard time finding tests for them.

MichaelClerx:
> It may sound silly but would you mind adding one or two small tests to capture the one DataLog line and the the realtime associated block in the LegacySimulation.run @MichaelClerx ? I have a hard time finding tests for them.

Sure!

PR #599, opened 2020-08-20
Sbml tweaks

MichaelClerx:
- moved myokit_model tests to bottom of sbml api test file
- rewrote a few tests to avoid parser
- updated compartment size test to add missing tests for rate
  - in general, these tests shouldn't test private methods, just the public API!
  - seems that setting rates for sizes wasn't covered in the tests? new test fails

### DIFF
--- a/myokit/formats/sbml/_api.py
+++ b/myokit/formats/sbml/_api.py
@@ -736,8 +736,15 @@ class Model(object):
             # Set initial value
             expr = compartment.initial_value()
             if expr is not None:
-                var.set_rhs(expr.clone(
-                    subst=expression_references))
+                try:
+                    var.set_rhs(expr.clone(
+                        subst=expression_references))
+                except AttributeError:
+                    raise SBMLError(
+                        'Initial value for the size of <' + str(compartment) +
+                        '> contains unreferenced parameters/variables. Please'
+                        ' use e.g. the `add_parameter` method to add reference'
+                        ' to parameters in the model.')
 
             if compartment.is_rate():
                 # Get initial state
--- a/myokit/formats/sbml/_api.py
+++ b/myokit/formats/sbml/_api.py
@@ -841,7 +841,13 @@ class Model(object):
                 expr = myokit.Multiply(conversion_factor, expr)
 
             # Get myokit amount variable
-            var = species_amount_references[species.sid()]
+            try:
+                var = species_amount_references[species.sid()]
+            except KeyError:
+                raise SBMLError(
+                    'Kinetic law of <' + str(reaction) + '> contains '
+                    'unreferenced species <' + str(species) + '>. Please, '
+                    'use the `add_species` method to reference species.')
 
             if not var.is_state():
                 # Get initial state
@@ -866,7 +872,14 @@ class Model(object):
                 expr = myokit.PrefixMinus(expr)
 
             # Set RHS
-            var.set_rhs(expr.clone(subst=expression_references))
+            try:
+                var.set_rhs(expr.clone(subst=expression_references))
+            except AttributeError:
+                raise SBMLError(
+                    'Reaction rate expression of <' + str(reaction) + '> for <'
+                    + str(species) + '> contains unreferenced parameters/'
+                    'variables. Please use e.g. the `add_parameter` method to '
+                    'add reference to parameters in the model.')
 
     def _set_rhs_products(
             self, reaction, variable_references, species_amount_references,
--- a/myokit/formats/sbml/_api.py
+++ b/myokit/formats/sbml/_api.py
@@ -1122,17 +1122,26 @@ class Model(object):
                 # Get sid
                 sid = species_reference.sid()
 
-                # Get stoichiometry variable
-                try:
-                    var = variable_references[sid]
-                except KeyError:
+                # If sid does not exist, there is no rhs to set
+                if sid is None:
                     continue
+
+                # Get stoichiometry variable
+                var = variable_references[sid]
 
                 # Set initial value
                 expr = species_reference.initial_value()
                 if expr is not None:
-                    var.set_rhs(expr.clone(
-                        subst=expression_references))
+                    try:
+                        var.set_rhs(expr.clone(
+                            subst=expression_references))
+                    except AttributeError:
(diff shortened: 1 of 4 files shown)


MichaelClerx:
@DavAug Hey David! Local test still showed two lines missing, added tests for them but they fail :D 

Could you have another look please? Also, tests should generally avoid testing private methods, because they shouldn't know how things work internally, just check that the exposed parts of the API do what they promise to do! But ok for now (not sure if they're needed though, if we just add tests that check all possibilities via the public api (initial value, normal value, ODE value)

DavAug:
> @DavAug Hey David! Local test still showed two lines missing, added tests for them but they fail :D
> 
> Could you have another look please? Also, tests should generally avoid testing private methods, because they shouldn't know how things work internally, just check that the exposed parts of the API do what they promise to do! But ok for now (not sure if they're needed though, if we just add tests that check all possibilities via the public api (initial value, normal value, ODE value)
> 
> See [68ea629](https://github.com/MichaelClerx/myokit/commit/68ea6293dae227b70a46ec82b7cb3c321ae115bb)

Haha great! Yes, I didn't notice var.eval() fails if var has a None rhs. So let me fix that...

DavAug:
@MichaelClerx 
Is it posible to set a states initial value to None? Otherwise I'd suggest we simply replace 

`var.promote(state_value=var.eval())`

by 

```
var.promote(state_value=var.eval() if var.rhs() else 0)
```

MichaelClerx:
> ```var.promote(state_value=var.eval() if var.rhs() else 0)```

this one please!

MichaelClerx:
> > `var.promote(state_value=var.eval() if var.rhs() else 0)`
> 
> this one please!

Uhm... Would this be an sbml error, by the way? Or a warning at least?

MichaelClerx:
> > > `var.promote(state_value=var.eval() if var.rhs() else 0)`
> > 
> > 
> > this one please!
> 
> Uhm... Would this be an sbml error, by the way? Or a warning at least?

Well, let's not bother. Just checked the CellML code, and I have the same there! (0 if init isn't set, and also RHS = 0 if not set).

It's not a good model, but at least it imports it then :D

DavAug:
> > > > `var.promote(state_value=var.eval() if var.rhs() else 0)`
> > > 
> > > 
> > > this one please!
> > 
> > 
> > Uhm... Would this be an sbml error, by the way? Or a warning at least?
> 
> Well, let's not bother. Just checked the CellML code, and I have the same there! (0 if init isn't set, and also RHS = 0 if not set).
> 
> It's not a good model, but at least it imports it then :D

Yes that's what I thought too. But I think according to the SBML documentation one has to set a intial state value, so for valid SBML files, this shouldn't be a problem. 

A warning could be good, but we can also just put this on the to do list.

MichaelClerx:
Similar one for the species now :D 
Not sure my expected results in the test are perfect, but problem I think is missing compartment size!

DavAug:
> Similar one for the species now :D
> Not sure my expected results in the test are perfect, but problem I think is missing compartment size!

Yes the problem is that the sbml.Compartment does not have an initial_value.

I think this should be an SBML error then? Because the model is not well defined for concentrations without defining the volume.

MichaelClerx:
> > Similar one for the species now :D
> > Not sure my expected results in the test are perfect, but problem I think is missing compartment size!
> 
> Yes the problem is that the sbml.Compartment does not have an initial_value.
> 
> I think this should be an SBML error then? Because the model is not well defined for concentrations without defining the volume.

Either that or we use default=1, similar to default initial condition=0 ?

DavAug:
Let’s make them all SBMLErrors?

On 20 Aug 2020, at 14:11, Michael Clerx <notifications@github.com> wrote:

﻿

Similar one for the species now :D
Not sure my expected results in the test are perfect, but problem I think is missing compartment size!

Yes the problem is that the sbml.Compartment does not have an initial_value.

I think this should be an SBML error then? Because the model is not well defined for concentrations without defining the volume.

Either that or we use default=1, similar to default initial condition=0 ?

MichaelClerx:
> Let’s make them all SBMLErrors?

That's fine by me!

I think in the CellML code I started out really strict (lots of errors raised), but then as I used the import more I started wanting to look at the contents of slightly-invalid models, so it became less strict again :D 

So up to you, really! Defaults, Defaults-with-warnings, Errors. All acceptable for me!

DavAug:
> > Let’s make them all SBMLErrors?
> 
> That's fine by me!
> 
> I think in the CellML code I started out really strict (lots of errors raised), but then as I used the import more I started wanting to look at the contents of slightly-invalid models, so it became less strict again :D
> 
> So up to you, really! Defaults, Defaults-with-warnings, Errors. All acceptable for me!

Hmm I guess defaults with warnings would be most desirable with this respect. Let me try to make those changes.

DavAug:
The are still one or two tests that fail, because I believe they are not quite correct. For example the rhs of the species concentration is not myokit.Number(4) is it?

MichaelClerx:
Feel free to fix the tests too :D

DavAug:
> Feel free to fix the tests too :D

Sure!
Just wanted to give you an update :D

DavAug:
@MichaelClerx 
Hey I added tests and handling of all edge cases I could come up with :D. The tests are not super tidy anymore, but I think it should be ok. Let me know what you think. :)

MichaelClerx:
Awesome!

MichaelClerx:
Looks great to me! Shall I merge it in?

DavAug:
> Looks great to me! Shall I merge it in?

Yes :)

DavAug (approved):

MichaelClerx:
Done! Version 1.31.0 includes all the new SBML work and is pip installable now! Thanks for all the hard work!

MichaelClerx:
Hope you're pleased with the results

DavAug:
Awesome! Thank you!

On Aug 26, 2020, at 5:10 PM, Michael Clerx <notifications@github.com<mailto:notifications@github.com>> wrote:



Done! Version 1.31.0 includes all the new SBML work and is pip installable now! Thanks for all the hard work!

PR #617, opened 2020-09-09
Old (abandoned) investigation into nmodl format.

MichaelClerx:
As used by NEURON

https://www.neuron.yale.edu/neuron/static/py_doc/modelspec/programmatic/mechanisms/nmodl.html

https://www.neuron.yale.edu/neuron/static/py_doc/modelspec/programmatic/mechanisms/nmodl2.html


MichaelClerx:
Language seems complicated, specific, and at least partially procedural

Issue #727, opened 2021-03-16
Make SubCapture thread safe


MichaelClerx:
SubCapture is used to hide compiler output, and doesn't look like there's another way to suppress that...

Users might run myokit in threads (shared memory) or multiprocessing (no shared memory), so need to make sure this doesn't cause any issues

MichaelClerx:
First test case:
```
#!/usr/bin/env python3
import myokit

x = myokit.SubCapture()
y = myokit.SubCapture()
z = myokit.SubCapture()

print('No capture')

x.enable()
print('x enabled')

y.enable()
print('y enabled')

z.enable()
print('z enabled')

y.disable()
print('y disabled')

x.disable()
print('x disabled')

z.disable()
print('z disabled')
```
Should only show 'No capture' and 'z disabled', but at the moment shows 'No capture', then 'x disabled', then nothing

MichaelClerx:
## Test case 2
No shared memory, starting new processes with `spawn`
This is how windows does it, and should lead to less resource-sharing
```
#!/usr/bin/env python3
import multiprocessing
import os
import time

import myokit


def pr(x):
    print(str(os.getpid()) + ': ' + str(x))


def f(t=0):
    pr('Running with t=' + str(t))
    with myokit.SubCapture():
        pr('Capturing')
        if t:
            pr('Sleeping')
            time.sleep(t)
            pr('Waking')
    pr('No longer capturing')


if __name__ == '__main__':
    multiprocessing.set_start_method('spawn')

    xs = [1, 5, 10]
    ps = []
    for x in xs:
        p = multiprocessing.Process(target=f, args=(x,))
        p.start()
        ps.append(p)
        time.sleep(0.5)

    for p in ps:
        if p.is_alive():
            p.join()
```
Example output:
```
[michael@laptop1 myokit]$ ./spawn.py 
WARNING:myokit:Using development version of Myokit (1.32.1.dev).
WARNING:myokit:Using development version of Myokit (1.32.1.dev).
12773: Running with t=1
WARNING:myokit:Using development version of Myokit (1.32.1.dev).
12797: Running with t=5
WARNING:myokit:Using development version of Myokit (1.32.1.dev).
12803: Running with t=10
12773: No longer capturing
12797: No longer capturing
12803: No longer capturing
```
and
```
[michael@laptop1 myokit]$ ./spawn.py 
WARNING:myokit:Using development version of Myokit (1.32.1.dev).
[...]

PR #728, opened 2021-03-16
Thread-safe output capturing

MichaelClerx:
See #727 

- [x] Backported cvode warning handling to legacy sim.
- [x] Restrict use of SubCapture to compilation
- [x] Thread-safe version of python capturing
- [x] Tests for python capturing
- [x] Thread-safe version of subprocess output capturing
- [x] Tests for subprocess output capturing


MichaelClerx:
Hi @martinjrobins!

I doubt you're still interested, but this is what I ended up with (see the changes to myokit/_aux).

It works a little different depending on whether you want "nested capturing" - which requires locking any other threads while you capture, or "interlaced capturing" - which only requires locking when applying/undoing the redirect.

For python stream capturing, I went with interlaced capturing, so you can do:

        a = StreamCapture()
        b = StreamCapture()
        a.start()
        print('This is captured by A')
        b.start()
        print('This is captured by A and B')
        a.stop()
        print('This is captured by B')
        b.stop()
        print('This is not captured')

this seems a bit weird, but makes more sense if you imagine a and b live in different threads that both write their output to the same stream, e.g. 

    def f(name):
      with StreamCapture():
        print('This is captured (at least) by ' + name)
    call f from two threads a and b

As you found, you'd get the interlaced entering and leaving here. So the above seemed the most natural solution.

I went a different route for the class that captures all output from this process and any C/C++ code etc. Here it was much harder to make multiple capturers capture the same output, as it deals with files (or file descriptors) instead of streams, so you need to read and write to the same file concurrently, which I didn't fancy trying. Also it has to be a temporary file but with a file descriptor, which isn't available by default so you'd have to manage creating/deleting that too.

So instead this one works in a nested manner: once you start capturing, no other threads are allowed to start capturing until you're done. You _are_ allowed capture twice in one thread, e.g.

        with StreamCapture() as a:
          print('This is captured by A')
          with StreamCapture() as b:
            print('This is captured by B but not A')
          print('This is captured by A')

Neither of these solutions should affect performance, because threading doesn't give a performance boost anyway. They both work fine if you're multiprocessing too, because then you don't share the streams.

Does that make any kind of sense?

I'm wondering now if I've overcomplicated it, and both capturers should be of the blocking/nesting type

MichaelClerx:
OK I think I've convinced myself now that the simpler nested approach is best in both cases 😄

MichaelClerx:
Closes #727

martinjrobins:
thanks for the update @MichaelClerx, agree that the nested approach sounds like a good universal solution :+1:

Issue #743, opened 2021-05-12
SimulationOpenCL.set_connections() does not seem to work with precision=64


ThomasMBury:
With precision=32, I am able to create arbitrary geometries in SimulationOpenCL using set_connections().

However, when I use precision=64, as has been required for some models, it seems to run without any connections i.e. no diffusion. No error messages appear.

I made a minimal working example attached below with 2 cells. Changing the precision from 32 to 64 breaks the diffusion between the two cells (at least for me!)
[mwe_test_connections.zip](https://github.com/MichaelClerx/myokit/files/6463417/mwe_test_connections.zip)

Any ideas @MichaelClerx?

MichaelClerx:
Thanks @ThomasMBury , especially for the example! 

Which version of Myokit is this?
Can you copy-paste the output of `python3 -m myokit system` below ?

ThomasMBury:
Sure, I was using the developer version:

WARNING:myokit:Using development version of Myokit (1.32.1.dev).
== System information ==
Myokit: 1.32.1.dev
Python: 3.9.4 (default, Apr  9 2021, 09:32:38) 
        [Clang 10.0.0 ]
OS: Darwin (darwin, posix)

== Python requirements ==
NumPy: 1.20.1
SciPy: 1.6.3
Matplotlib: 3.4.2
ConfigParser: OK
Setuptools: 52.0.0.post20210125

== Python extras ==
SymPy: Not found
CMAES: Not found
MoviePy: Not found

== GUI ==
PyQt5: 5.9.7
  Sip: OK
PyQt4: Not found
PySide: Not found
PySide2: Not found

== Development tools ==
Sphinx: 3.5.3
Flake8: 3.9.0

== Simulation tools ==
Compiler: Clang (Apple LLVM) 12.0.0
Sundials: 5.6.1
OpenCL: 2 device(s) found
  Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz on Apple
  Intel(R) Iris(TM) Plus Graphics 645 on Apple
  Use `python -m myokit opencl` for more information.

= /Users/tbury/.config/myokit/myokit.ini =
[myokit]
# This file can be used to set global configuration options for Myokit.

[time]
# Date format used throughout Myokit
# Format should be acceptable for time.strftime
date_format = %Y-%m-%d %H:%M:%S
# Time format used throughout Myokit
time_format = %H:%M:%S

[debug]
# Add line numbers to debug output of simulations
line_numbers = True

[...]

PR #744, opened 2021-05-15
OpenCL sim connections with double precision

MichaelClerx:
See #743 

The atomic add operation for floating points only works on floats. Replaced with a version that (1) fixes this, and (2) avoids another possible problem (see https://streamhpc.com/blog/2016-02-09/atomic-operations-for-floats-in-opencl-improved)


MichaelClerx:
OS/X tests are currently failing with quite an obscure error: CVMS_ERROR_INVALID_CONNECTION

However, this does not seem related to these fixes, as we saw the same in https://github.com/MichaelClerx/myokit/pull/746

Issue #665, opened 2020-11-13
Allow components from other models to be imported


MichaelClerx:
Proposed method `myokit.Model.import_component`:

```
def import_component(
        self, external_component, new_name=None, var_map=None, 
        allow_name_mapping=False, convert_units=False):
    """
    Imports a component from another model (the "source model") into this one 
    (the "target model").

    If variables in the ``external_component`` refer to variables from other 
    components they will be mapped on to variables from this model by trying
    the following strategies (in order):

    1. Using the provided dict ``var_map``, which maps variables in the source
       model to variables in this model.
    2. By assuming variables with the same label/binding map onto each other.
    3. By assuming that variables with the same qualified names map onto each
       other. This rule is only applied if ``allow_name_mapping`` is set to
       ``True``.

    Arguments:
    
    ``external_component``
        A :class:`myokit.Component` from another model.
    ``new_name``
        An optional new name for the imported component.
    ``var_map``
        An optional dict mapping variables from the source model to variables
        in this model (with variables specified as objects or as fully
        qualified names).
    ``allow_name_mapping``
        Set this to ``True`` to allow mapping based on fully qualified names,
        e.g. a variable ``x.y`` in the source model will be mapped to a
        variable ``x.y`` in the target model.
    ``convert_units``
        Set this to ``True`` to convert the units of any mapped variables, if
        required and possible.

    If any variables cannot be mapped, a :class:`myokit.WellNamedError`
    will be raised. Will raise a :class:`DuplicateNameError`(?) if the 
    components name or ``new_name`` clashes with an existing component name. If
    unit conversion is enabled and variables with incompattible units are
    mapped onto each other, a :class:`myokit.IncompatibleUnitsError` will be
    raised.    
    """
    
    # Check component is not from this model, is a component, etc.
    
    # Check if new name is provided, or else check that name doesn't clash
    
    # Create mapping for vars, and get unit conversion factors if required.
    
    # Clone contents of old component to new one
    # Not sure what the best strategy is. Re-write based on code from clone()?
    # Might need hidden methods in Component and in Variable.

    ...
```

MichaelClerx:
Would this be enough, @Rebecca-Rumney @DavAug ?

MichaelClerx:
(One tiny issue you'd have to deal with is user functions. Hardly ever used, but they exist! Basically you'd have to either copy the user function or (safer) rewrite it as an expression during the copying, probably just 3 lines + some tests)

[...]

PR #769, opened 2021-06-07
Importing components i665

Rebecca-Rumney:
Solving issue #665 .


MichaelClerx:
Thanks @Rebecca-Rumney !

Did you try in on your PKPD code, and did it solve your problems?

Any unexpected difficulties or things I should know about before reviewing?

Rebecca-Rumney:
@MichaelClerx Yes, it did work on my PKPD code. It was slightly more complicated than I initially thought it would. As I'm importing multiple components that are interconnected, I have to instantiate the variables in a new temporary component before starting the import.

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+
+        ext_model_copy = external_component.model().clone()
+        external_component_copy = ext_model_copy[external_component.qname()]
+
+        full_var_map = {}
+        relevant_vars = external_component_copy.variables()
+        for var in external_component_copy.variables():
+            dependent_vars = var.refs_to(state_refs=False)
+            dependent_states = var.refs_to(state_refs=True)
+            relevant_vars = (
```
MichaelClerx, line 1735: This creates a new tuple every iteration. Better to use `relevant_vars.extend(var.refs_to())`

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                list(relevant_vars) +
+                list(dependent_vars) +
+                list(dependent_states)
+            )
+
+        try:
```
MichaelClerx, line 1741: It's probably a good idea to check if both models are valid at the start of this method, e.g.

        if not model.is_valid():
            model.validate()
            
Then you can just use .time() without the try-catch because a valid model always has a time variable

MichaelClerx: Only downside is it then won't work for invalid models, but that's probably something we can live with?

myokit/_model_api.py
MichaelClerx: Since you're making a set later, better still to use `relevant_vars = set()` above, and then keep adding to it with `relevant_vars.update(var.refs_to())`

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                        raise myokit.WellMappedError(
+                            self_var + ' does not exist in this model'
+                        )
+                else:
+                    raise TypeError(
+                        'Variables in the var_map need to be specified as ' +
+                        'objects or as fully qualified names'
+                    )
+
+                # check external variables in var_map are variables and exists
```
MichaelClerx, line 1776:
```suggestion
                # check external variables in var_map are variables and exist
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+        if var_map is not None:
+            if not isinstance(var_map, dict):
+                raise TypeError('var_map needs to be of type dict or None')
+            for ext_var, self_var in var_map.items():
+                # check self variables in var_map are variables and exists
+                # in self
+                if isinstance(self_var, myokit.Variable):
+                    if not self_var.has_ancestor(self):
+                        raise myokit.WellMappedError(
+                            self_var.name() + ' does not exist in this model'
```
MichaelClerx, line 1761:
```suggestion
                            self_var.qname() + ' does not exist in this model'
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                        'Variables in the var_map need to be specified as ' +
+                        'objects or as fully qualified names'
+                    )
+
+                # check external variables in var_map are variables and exists
+                # in external model
+                if isinstance(ext_var, myokit.Variable):
+                    if not ext_var.has_ancestor(external_component.model()):
+                        raise myokit.WellMappedError(
+                            ext_var.name() +
```
MichaelClerx, line 1781:
```suggestion
                            ext_var.qname() +
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                            ' does not exist in the external component\'s' +
+                            'model'
```
MichaelClerx, line 1783:
```suggestion
                            ' model'
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                        )
+                    ext_var = ext_model_copy.var(ext_var.qname())
+                elif isinstance(ext_var, basestring):
+                    try:
+                        ext_var = ext_model_copy.var(ext_var)
+                    except KeyError:
+                        raise myokit.WellMappedError(
+                            ext_var +
+                            ' does not exist in the external component\'s' +
+                            'model'
```
MichaelClerx, line 1793:
```suggestion
                            ' model'
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+        if new_name is None:
+            new_name = external_component.name()
+        if self.has_component(new_name):
+            raise myokit.DuplicateName(
+                'This model already has component with the name ' + new_name
+            )
+
+        # Create mapping for vars, and get unit conversion factors if required.
+
+        ext_model_copy = external_component.model().clone()
```
MichaelClerx, line 1727: Why the clones? (This is quite expensive)

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+            relevant_vars.append(ext_model_copy.timex())
+        except myokit.IncompatibleModelError:
+            pass
+
+        relevant_vars = list(set(relevant_vars))
+        # print(relevant_vars)
+
+        copied_var_list = []
+
+        # map has form {external_model_copy.variable: self.variable}
```
MichaelClerx, line 1751: Add some comment saying "Check user-specified var map"

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                if self_var in copied_var_list:
+                    raise myokit.WellMappedError(
+                        'Multiple variables map onto ' + self_var.name()
+                    )
+
+                # add to full_var_map if a relevant variable
+                if ext_var in relevant_vars:
+                    full_var_map[ext_var] = self_var
+                copied_var_list.append(self_var)
+
```
MichaelClerx, line 1811: Add some comment saying what the next bit does

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                    if self.has_variable(ext_var.qname()):
+                        # add to full map
+                        full_var_map[ext_var] = self.get(ext_var.qname())
+            if (
+                ext_var not in full_var_map and
+                ext_var not in external_component_copy.variables(deep=True)
+            ):
+                raise myokit.WellMappedError(
+                    ext_var.qname() +
+                    ' is refered to by the external_component but has no' +
```
MichaelClerx, line 1835:
```suggestion
                    ' is referred to by the external_component but has no'
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                    'mapping in this model'
```
MichaelClerx, line 1836:
```suggestion
                    ' mapping in this model'
```

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                    # add self_bind_var:ext_var to full_var_map
+                    full_var_map[ext_var] = self_bind_var
+                elif self_label_var is not None:
+                    # add self_label_var:ext_var to full_var_map
+                    full_var_map[ext_var] = self_label_var
+                elif allow_name_mapping:
+                    if self.has_variable(ext_var.qname()):
+                        # add to full map
+                        full_var_map[ext_var] = self.get(ext_var.qname())
+            if (
```
MichaelClerx, line 1829: What about the name mapping?

myokit/_model_api.py
```diff
@@ -1659,6 +1659,276 @@ def has_warnings(self):
...
+                # variable
+                ext_var.convert_unit(self_var.unit())
+
+        # create empty component in self
+        new_component = self.add_component(new_name)
+        external_component_copy._clone_modelpart_data(new_component)
+        ext_temp_component = ext_model_copy.add_component_allow_renaming(
+            new_name
+        )
+
```
MichaelClerx, line 1851: Add a comment saying what the next bit does

MichaelClerx (changes requested):
Hi @Rebecca-Rumney !

Added some minor comments on the first half of the code. I get a bit lost when the actual variable creation happens. Why do we need a clone of the external model?

myokit/_model_api.py
Rebecca-Rumney: I was thinking there might be users who want to create a full model out of incomplete models and so wouldn't have fully valid models to input. Is this bit of code problematic or is it acceptable to leave like this to allow invalid models?

myokit/_model_api.py
Rebecca-Rumney: The clone is because I need to change some things about the external model (converting units and altering the structure of the model so the equations reference the correct variables). I do realise this is quite expensive but that was the easiest way I could find to do it without changing the external model.

build/lib/myokit/_model_api.py
```diff
@@ -0,0 +1,4623 @@
+#
```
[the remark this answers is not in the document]
Rebecca-Rumney: Yep sorry. Now deleted.

myokit/_err.py
```diff
@@ -418,3 +418,10 @@ def __init__(self, var):
         tok = var.lhs()._token
         super(UnusedVariableError, self).__init__(msg, tok)
 
+
+class WellMappedError(MyokitError):
```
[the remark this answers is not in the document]
Rebecca-Rumney: Yes, that is a much better name.

myokit/_model_api.py
Rebecca-Rumney: Thanks, I didn't realise it was creating tuples. Code updated

myokit/_model_api.py
Rebecca-Rumney: comments have been updated and added throughout, hopefully it's more readable now.

myokit/_model_api.py
Rebecca-Rumney: The name mapping is there. lines 1825 to 1828
```
    elif allow_name_mapping:
        if self.has_variable(ext_var.qname()):
            # add to full map
            full_var_map[ext_var] = self.get(ext_var.qname())
```

myokit/_model_api.py
MichaelClerx: Good point!

myokit/_model_api.py
Rebecca-Rumney: I have now implemented the function in a way that does not require cloning the model.

MichaelClerx:
Hi @Rebecca-Rumney !

During reviewing, I realised you could do the unit conversion more easily by passing in unit-converted expressions to the `lhsmap` argument of the `_clone` method. There were also a couple of edge-cases I was worried about (e.g. units of derivatives) so I went and added a bunch of things.

Can you check if this code still does what you need it to do? If so I'm happy to merge it in.

PR #868, opened 2022-04-20
Offline tests for memory leaks

MichaelClerx:
Not easy


MichaelClerx:
@DavAug I tried to do some automated testing but it's remarkably hard, so have added some scripts to help us debug if this happens again / somewhere else

Issue #870, opened 2022-04-21
Memory leak with sensitivities


DavAug:
Very similar to #867 , there seem to be residual memory allocations from the sensitivities.

Toy problem:
```
import resource
import myokit

m, p, _ = myokit.load('example')

# Enable sensitivities
states = [var.qname() for var in m.states()]
params = []
for var in m.variables(const=True):
    # Sometimes constants are derived from parameters
    if not var.is_literal():
        continue
    params.append(var.qname())
s = myokit.Simulation(m, p, (states, params))

s.run(10, log=myokit.LOG_NONE)

u1 = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
for i in range(100):
    s.run(1, log=myokit.LOG_NONE)
u2 = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss

print(u2 - u1)
```

Without sensitivities this returns ``0`` (as we would expect). With sensitivities this returns 

```
(venv) ➜  myokit git:(main) ✗ python todelete.py
WARNING:myokit:Using development version of Myokit (1.33.4.dev).
311296
```

MichaelClerx:
> @MichaelClerx Is it possible that
> 
> ```
> sy = N_VCloneVectorArray(model->ns_independents, y);
> ```
> 
> and
> 
> ```
> if (sy != NULL) { N_VDestroyVectorArray(sy, model->ns_independents); sy = NULL; }
> ```
> 
> are not counterparts of each other?

It's possible, but you can check that in the sundials documentation :D
I think it's correct though, and we would get an error otherwise

MichaelClerx:
I've added a slightly more complicated test script here: https://github.com/MichaelClerx/myokit/tree/main/myokit/tests/memory
And just added your testcase (sort of). With the default settings it shows no change. If I make it more like your toy problem (protocol from example file, no s.reset(), same run durations) I get a single memory increase. For definite memory leaks I usually saw 2 or 3 (try it by commenting out a free() line).

I have no idea why this seems to be non-deterministic 🤷🏻‍♂️ 

Maybe a good idea to also ask Martin or Fergus what they think is the best way to find these memory leaks? (And check that they actually are memory leaks, not e.g. some Python list somewhere increasing in size).

DavAug:
> > @MichaelClerx Is it possible that
> > ```
> > sy = N_VCloneVectorArray(model->ns_independents, y);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
[...]

PR #883, opened 2022-07-27
Memory leak when using `log_times`

MichaelClerx:
Only in cvodessim, not in cvode sim.
Introduced when switching from list to sequence interface for `log_times` handling

log_times

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This page lists the main changes made to Myokit in each release.
 - Deprecated
 - Removed
 - Fixed
+  - [#883](https://github.com/MichaelClerx/myokit/pull/883) Fixed memory leak in `myokit.Simulation` when using the `log_times` argument.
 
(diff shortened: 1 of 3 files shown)


MichaelClerx:
Hey @DavAug , I just found and fixed a much larger memory leak than we were seeing in #870.
It occurs only when you use the `log_times` argument, which I guess you were doing in your real fits?

By comparison, I was losing nearly 10mb per simulation on this. Compared to 0.03mb per simulation as seen in the graphs for #870 

Do you think this is what you were seeing?

DavAug:
Wow very good spot! Yes, I was logging the times, so this explains why my inference consumed so much memory 😄

MichaelClerx:
Sorry for not figuring it out earlier @DavAug ! I was really surprised when we wrote the test and it didn't show a leak. But not surprised enough to figure out that the test was wrong, apparently 😓

Issue #887, opened 2022-08-02
conversion from g to nmol


martinjrobins:
I want to use `conversion_factor(unit1, unit2, helpers=None)` to convert between [g] to [nmol]. I'm guessing that I can use `helpers` in some way by specifying the molecular weight of the compound, but the documentation for this arg is not clear. How can I do this?

MichaelClerx:
You would have to provide a `Quantity` (or a string representation) to convert from g to nmol, so e.g. `"100 [g/nmol]"`.
The interface assumes a list of helpers, so it'd have to be e.g. `["100 [g/nmol]"]`, or `[myokit,Quantity(100, 'g/nmol')]`or something like that.

Does that help? I'm very happy to discuss ways to improve the docs.

(Also, the `Quantity` class is pretty obscure, and it's on the to-do list to move all its functionality into `Number`: https://github.com/myokit/myokit/issues/798 )

martinjrobins:
Yes, that helps, a couple of questions:
1. Does the equation `[unit1] * c = 1 [unit2]` still hold for the returned conversion factor `c`? i.e. is the helper incorporated into `c`?
2. Is `c` independent of the units that I give for the helper. i.e. what if I provide `1e-6 [g/mol]`
3. Why can I provide a list of helpers, what is the advantage here?

MichaelClerx:
>     1. Does the equation `[unit1] * c = 1 [unit2]` still hold for the returned conversion factor `c`? i.e. is the helper incorporated into `c`?

Yup! 

>     2. Is `c` independent of the units that I give for the helper. i.e. what if I provide `1e11 [g/mol]`, or even `1e-2 [nmol/g]`?
Did you look at the code or just guess this? :D 
Yes those are exactly the 2 additional cases it handles: the helper is scaled as necessary and possibly even inverted.

>     3. Why can I provide a list of helpers, what is the advantage here?

This just makes it easier to convert several variables with the same syntax. E.g. 

```
helpers=[case1, case2, case3]
for unit in list_including_case_1_2_and_3:
    convert(unit, helpers)
```

MichaelClerx:
Do you think the updates in https://github.com/myokit/myokit/pull/888 are enough to clarify this for future users?

martinjrobins:
Thanks @MichaelClerx, that is great :) I've added a review on #888 (all looks good) and can close this now

MichaelClerx:
Thanks @martinjrobins !
I can't believe you found this feature 😆

[...]

PR #888, opened 2022-08-03
Clarify docs for Unit.conversion_factor with "helpers"

MichaelClerx:
See #887


martinjrobins (approved):
Thanks @MichaelClerx, I think this is much clearer 👍

Issue #894, opened 2022-08-23
CVODES sim with time-dependent variables doesn't compile


MichaelClerx:
```
#!/usr/bin/env python
import myokit

model = myokit.parse_model('''
[[model]]
c.x = 0.5

[c]
time = 0 bind time
dot(x) = (1 - x) / 1000
V = time
g = 1
i = g * V
''')

p = ['c.g']
i = ['c.i']
myokit.Simulation(model, sensitivities=(i, p))
```

```
    /tmp/tmpuc9hddgvmyokit/source.c: In function ‘Model_EvaluateSensitivityOutputs’:
    /tmp/tmpuc9hddgvmyokit/source.c:797:32: error: ‘S0_V_V’ undeclared (first use in this function); did you mean ‘S0_V_i’?
      797 |     S0_V_i = 1.0 * V_V + P_g * S0_V_V;
          |                                ^~~~~~
          |                                S0_V_i
    /tmp/tmpuc9hddgvmyokit/source.c:797:32: note: each undeclared identifier is reported only once for each function it appears in
```

PR #895, opened 2022-08-23
Added test for issue 894 with derivs involving bound variables

MichaelClerx:
See #894.


myokit/_expressions.py
```diff
@@ -187,27 +187,67 @@ def depends_on(self, lhs, deep=False):
...
+        # Shallow check
+        if not deep:
+            for lhs in self._references:
+                if lhs.is_state_value():
+                    return True
+            return False
+
+        # Determine if this variable has a (deep) dependency on a state
+        todo = set(self._references)
+        done = set([self])
```
MichaelClerx, line 233:
```suggestion
        done = set()
```

Issue #435, opened 2019-12-02
Change `state_value()` and `set_state_value` to `initial_value` ?


MichaelClerx:
Will make it easier to find

MichaelClerx:
Unchanged:
- model.states() --> returns list of state variables

Deprecated:
- [ ] model.state() --> returns list of floats
- [ ] model.inits() --> returns list of equations
- [ ] variable.state_value() --> returns float
- [ ] variable.set_state_value() --> sets initial value as float

New:
- [ ] model.initial_state(as_floats=False, as_equations=False) --> returns expressions, floats, or equations
- [ ] variable.set_initial_state --> set initial state as float or expr
- [ ] variable.initial_state(as_float=False) --> return expr or float

MichaelClerx:
Closed in #918

PR #899, opened 2022-10-12
I898 initial expression

### LINKED ISSUE #898: state variable with an initial condition dependent on another (constant) variable
I'm trying to define a state variable in a myokit model, that has a default state that is an expression containing other constant variables. Is this possible? From the docs I suspect it is not, and the default state must be a concrete value?

martinjrobins:
Fixes #898 

variable initial values are now myokit Expressions, must be constant

**edit by michael, 2023-02-22**

Remaining to-do:

**Simulations**
- [x] `Simulation`: Convert. Add tests to check that conversion happens.
- [x] `Simulation`: Warn if indep var is a constant that's referenced (even indirectly) by an initial value.
- [x] `Simulation1d`: Convert. Add tests to check that conversion happens.
- [x] `SimulationOpenCL`: Convert. Add tests to check that conversion happens.
- [x] `FiberTissueSimulation`: Convert. Add tests to check that conversion happens.
- [x] `RhsBenchmarker`: Convert.
- [x] `JacobianTracer`: Convert.
- [x] `JacobianCalculator`: Convert.
- [x] `lib.hh` Convert. Add tests.
- [x] `lib.markov` Convert. Add tests.

**Exports**
- [x] CellML v1 export: Reduce to number.
- [x] CellML v2 export: Reduce to number.
- [x] Euler export: Reduce to number.
- [x] CVODE export: Reduce to number.
- [x] Cable export: Reduce to number.
- [x] CUDA export: Reduce to number.
- [x] OpenCL export: Reduce to number.
- [x] HTML export: Doesn't show initial values.
- [x] XML export: Doesn't show initial values.
- [x] Latex export: Leave in.
- [x] EasyML: Reduce to number.
- [x] Matlab: Reduce to number.
- [x] Python: Reduce to number.
- [x] Stan: Reduce to number.


MichaelClerx:
Hey @martinjrobins , glad to see this moving ahead. Just checking: Do you need expressions to stay part of the current state? So not just a parse-time construct?

martinjrobins:
> Hey @martinjrobins , glad to see this moving ahead. Just checking: Do you need expressions to stay part of the current state? So not just a parse-time construct?

yea, I need to keep the expression since the parameters that this expression depends on might change

martinjrobins:
Hi @MichaelClerx, I think this is done now, I just need to cover a couple of lines (for some reason codecov is not telling me which ones!). Be good to get a review from you

martinjrobins:
actually, I should also add a test of restarting a simulation with an initial state that is an expression and check that it changes. I'll do that once I've got a review to check that I'm on the right track

MichaelClerx:
Thanks @martinjrobins ! This is a sit-down-and-think one, so might be a while, but happy to make some time

martinjrobins:
Thanks @MichaelClerx. I realise that there is more to do on the simulations side to verify that all the simulation and code generation work, but I just want to check that you're happy with the changes to Model and parser before I do this

myokit/__init__.py
```diff
@@ -359,6 +359,7 @@
     UnresolvedReferenceError,
     UnusedVariableError,
     VariableMappingError,
+    NonConstantExpressionError,
```
MichaelClerx, line 362: I'm doing these alphabetically

myokit/_err.py
```diff
@@ -426,3 +426,11 @@ class VariableMappingError(MyokitError):
 
     *Extends:* :class:`myokit.MyokitError`.
     """
+
+
+class NonConstantExpressionError(MyokitError):
```
MichaelClerx, line 431: These are also alphabetical (where possible)

myokit/_model_api.py
```diff
@@ -2933,7 +2933,9 @@ def set_state(self, state):
         Changes this model's state. Accepts any type of input handled by
         :meth:`map_to_state`.
         """
-        self._current_state = self.map_to_state(state)
+        self._current_state = [
+            myokit.Number(x) for x in self.map_to_state(state)
```
MichaelClerx, line 2937: It **might** make sense to update `map_to_state` to returns Number objects.

MichaelClerx: Quick grep:
```
./myokit/_sim/cvodesim.py:            y = self._model.map_to_state(y)
./myokit/_sim/cvodesim.py:        self._default_state = self._model.map_to_state(state)
./myokit/_sim/cvodesim.py:        self._state = self._model.map_to_state(state)
./myokit/_sim/cvodessim.py:            y = self._model.map_to_state(y)
./myokit/_sim/cvodessim.py:        self._default_state = self._model.map_to_state(state)
./myokit/_sim/cvodessim.py:        self._state = self._model.map_to_state(state)
./myokit/_io.py:            s = model.map_to_state(s)
./myokit/_io.py:        state = model.map_to_state(state)
./myokit/_model_api.py:            self.set_state(self.map_to_state(state))
./myokit/_model_api.py:        self._current_state = self.map_to_state(state)
./myokit/__main__.py:        ref = model.map_to_state(ref)
```

So maybe not

myokit/_model_api.py
```diff
@@ -2055,7 +2055,7 @@ def inits(self):
         def StateDefIterator(model):
             for k, var in enumerate(model._state):
                 yield Equation(
-                    myokit.Name(var), myokit.Number(self._current_state[k]))
+                    myokit.Name(var), self._current_state[k])
```
MichaelClerx, line 2058: Probably fits on one line now?
Esp if we switch to `for var, state in zip(model._state, self._current_state)`

myokit/_model_api.py
```diff
@@ -3256,7 +3258,7 @@ def state(self):
         Returns the current state of the model as a list of floating point
         numbers.
         """
-        return list(self._current_state)
+        return [float(eqn.rhs) for eqn in self.inits()]
```
MichaelClerx, line 3261: This is definitely the path of least resistance. But should we also provide users with a way to get the real state, expressions & all ?

Even without redefining the return type, we'll need to do a grep for all calls in Myokit to `state()`, and check that they get what they are expecting (i.e. did they want floats of the current state or does it make more sense for them to get expressions now that that's an option)

myokit/_model_api.py
```diff
@@ -4357,10 +4359,11 @@ def lhs(self):
         """
         return self._lhs
 
-    def promote(self, state_value=0):
+    def promote(self, state_value=0, check_const=True):
```
MichaelClerx, line 4362: I think I'd prefer to leave this to the validation stage? So have the checks in `validate` only.

I've tried a few times to make one of those APIs that doesn't let you construct a bad model, but it makes it very tiresome to use!

myokit/_model_api.py
```diff
@@ -4357,10 +4359,11 @@ def lhs(self):
...
         """
         Turns this variable into a state variable with a current state value
-        given by ``state_value``.
+        given by ``state_value``. state_value could be a value, an Expression,
+        or a proto expression
```
MichaelClerx, line 4366: Oh no! You are the first person ever to discover the dirty secret that is proto expressions 🙈 

Maybe omit them here for clarity? A user using Myokit in a non-hacky way shouldn't have to know about them, I think.

myokit/_model_api.py
```diff
@@ -4744,7 +4753,7 @@ def state_value(self):
         """
         if not self._is_state:
             raise Exception('Only state variables have initial values.')
-        return self.model()._current_state[self._indice]
+        return float(self.model()._current_state[self._indice])
```
MichaelClerx, line 4756: As with `state`, the user will need some way to access the real state value

myokit/_model_api.py
martinjrobins: agree maybe leave it for now?

myokit/_model_api.py
martinjrobins: agreed. Shall I define another function `state_values()` that returns the list of numbers, and then have `state()` return the equations themselves? Any existing calls to state() I can decide to either use `state()` or `state_values()` depending on what is appropriate

myokit/_model_api.py
MichaelClerx: yeah

myokit/_model_api.py
martinjrobins: ok, so in doing this I've come across a problem regarding `set_state`. This is used to set the current state of the model to do some calculations, e.g. `evaluate_derivatives`. Presumable the intent here is that you can set the current state to something that might not be the initial state? So do we need `self._current_state`, a list of floats, and `self._initial_state`, a list of expressions? So you are free to change the current state independently from the initial_state?

MichaelClerx: The model is only supposed to have a single state (which I might want to rename initial state at some point #435). I think evaluate_derivatives only changes it as a bit of a hack (would have been in a try-finally block if I programmed this again!)
I think the answer here is for org_state to be the state-with-variables, while the subsequent set_state uses floats, and then at the end set_state(state-with-variables) is used again to restore the original state

MichaelClerx: An even better solution might be to rewrite the whole thing to use `subst={Name(state_x): value_x, ...}` in the calls to eval().

MichaelClerx: Shall I make a ticket for that and rewrite it? Then you can pull that in

martinjrobins:
question about `format_state`, this formats as a list of numbers, do you still want this behaviour? Its used in `save_state` to save to a file. Not sure why this is needed as the initial states are in the model code arn't they?

MichaelClerx:
> question about `format_state`, this formats as a list of numbers, do you still want this behaviour? Its used in `save_state` to save to a file. Not sure why this is needed as the initial states are in the model code arn't they?

Yeah definitely! When porting models from original code in e.g. matlab or C or fortran I often check the derivatives at a handful of initial states against the original. This involves some loading and saving to disk of different states

myokit/_model_api.py
martinjrobins: so I've left in the constant check, as I think it is needed if a user uses promote and passes a non-constant expression for `state_value`

myokit/_model_api.py
martinjrobins: done

myokit/_model_api.py
martinjrobins: but I've removed the `check_const` flag, as now the parser calls `promote` without `state_value` (i.e. now promote does not support proto expressions)

martinjrobins:
@MichaelClerx: how do you want the simulation classes to behave? I'm thinking that `_default_state` would copy the expressions for the initial state from the model initially (afterwards the user could overwrite as they want). When the simulation is `reset()`, the `_state` is updated from the `_default_state` like so:

```
self._state = [float(expr) for expr in self._default_state]
```

So if you wanted to change a constant in the model, and this affected the initial state, all you have to do is:

```python
        m = myokit.Model()
        c = m.add_component('c')
        t = c.add_variable('t')
        t.set_rhs(0)
        t.set_binding('time')
        p = c.add_variable('p')
        p.set_rhs('1')
        y = c.add_variable('y')
        y.promote(myokit.Name(p))
        y.set_rhs('-y')

        s = myokit.LegacySimulation(m)
        d = s.run(1)
        self.assertAlmostEqual(s.state()[0], 1 * np.exp(-1), 3)

        s.set_constant('c.p', 2)
        s.reset()
        d = s.run(1)
        self.assertAlmostEqual(s.state()[0], 2 * np.exp(-1), 3)
```

MichaelClerx:
@martinjrobins that's making my head hurt a bit!

I think what you're saying makes sense, yes: the state is floats, but the default state is expressions.

That's a thing that might break some user code, I guess, if they are doing things like `sim1.set_default_state(sim2.state())` --- unless we have good conversion in set_[default_]state

martinjrobins:
> @martinjrobins that's making my head hurt a bit!
> 
> I think what you're saying makes sense, yes: the state is floats, but the default state is expressions.
> 
> That's a thing that might break some user code, I guess, if they are doing things like `sim1.set_default_state(sim2.state())` --- unless we have good conversion in set_[default_]state
 
yup, I've put in a conversion in `set_default_state` that handles this, and will do the same in `set_state` in case expressions are passed in

MichaelClerx:
Sounds really good! If we can get it so that models that don't use expressions in the initial state (and code for those models) still work as expected, then that would be a very good result.

martinjrobins:
hi @MichaelClerx. All simulators look good apart from the opencl one, getting a python 2.7 only bug but cannot seem to install python 2.7 on my laptop, if you have any time do you mind having a quick look? (if not no worries, I'll come back to it next week and have another go)

MichaelClerx:
```
            x, y = 2, 1
            sx = [myokit.Number(i) for i in list(range(m))]
            self.s2.set_state([0] * m)
            self.s2.set_default_state(sx, x, y)
```
It was these lines! In Python 2.7 the leak comprehension "leaks" its x, so that x is m-1=7 when set_default_state is called

martinjrobins:
thanks @MichaelClerx, nice catch I didn't realise that this could happen!

martinjrobins:
@MichaelClerx : I think this is all done now? I had one question re cyclic dependencies, I think we don't have to worry about this as an initial state can only depend on constant variables, so can't have any cycles, do you agree?

Other than that it is ready for review.

MichaelClerx:
Thanks @martinjrobins ! What was the cyclic question? I think you can have cycles in networks of constants, just not with literals. (Although it depends on exactly where in the chain of events the "is_constant" flag gets set: if you have a = b and b = a then that might never happen?)

martinjrobins:
It was something that you brought up in the original issue. But a constant variable can't depend on an initial condition right, I don't think you can even write that in .mmt format?

MichaelClerx:
> It was something that you brought up in the original issue. But a constant variable can't depend on an initial condition right

Good point!

> I don't think you can even write that in .mmt format?

Surprisingly, there is syntax for that but it's not allowed in a normal model, only when you are selecting partial derivatives to return from a cvodes sim :D :D :D 

I will try to review before the week is over

myokit/_model_api.py
```diff
@@ -4356,15 +4370,18 @@ def promote(self, state_value=0):
...
-                    'Expressions for state values can not contain references'
-                    ' to other variables.')
+        if not isinstance(state_value, myokit.Expression):
+            state_value = myokit.Number(state_value)
 
         model = self.model()
+
+        # check initial state value expression is constant
+        if not state_value.is_constant():
+            raise myokit.NonConstantExpressionError(
```
MichaelClerx, line 4380: I think we can delete the NonLiteralValueError then?

myokit/_err.py
```diff
@@ -292,6 +292,14 @@ def __init__(self):
         super(MissingTimeVariableError, self).__init__(msg)
 
 
+class NonConstantExpressionError(MyokitError):
```
MichaelClerx, line 295: Please note in changelog under "Added"

myokit/_model_api.py
```diff
@@ -4344,7 +4357,8 @@ def lhs(self):
     def promote(self, state_value=0):
         """
         Turns this variable into a state variable with a current state value
-        given by ``state_value``.
+        given by ``state_value``. state_value could be a value or an
+        Expression,
```
MichaelClerx, line 4361:
```suggestion
        given by ``state_value``.
        
        The new ``state_value`` should be either a numerical value or a
        :class:`myokit.Expression`. If an expression is used, it can contain
        references to model variables, as long as they are constant-valued.
```

something like that

myokit/tests/test_model.py
```diff
@@ -2079,6 +2079,22 @@ def test_validate_and_remove_unused_variables(self):
         m.validate()
         self.assertEqual(len(m.warnings()), 0)  # issue fixed!
 
+    def test_state(self):
```
MichaelClerx, line 2082: Needs a tiny comment

But also needs to test state()?

myokit/tests/test_parsing.py
```diff
@@ -1404,6 +1418,19 @@ def test_initial_values(self):
             """
         self.assertRaises(myokit.ParseError, myokit.parse, code)
 
+        # non-constant initial value
```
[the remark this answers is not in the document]
MichaelClerx: (Sorry, doesn't have to be in the parser test of course, as long as its somewhere)

MichaelClerx (changes requested):
Hi @martinjrobins ! I think you've covered most bases. Couple of suggestions and a few more things that spring to mind:

MichaelClerx:
1. Should initial values be allowed depend on child variables? The usual rule is that you can only access your own child vars, not anybody elses. Not sure if I feel like the init cond. belong to the variable or to the whole model. What's it like in your use cases?

MichaelClerx:
2. Can we get a test where we promote so that `init(x) = a; a = b; b = 1 + a`? This should be ignored by promote(), but state_values should raise an error (e.g. by first calling validate()) instead of getting caught in some endless loop
3. Similarly, if we create the same mechanism via a set_rhs on a or b

myokit/tests/test_parsing.py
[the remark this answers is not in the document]
MichaelClerx: an `a.b = dot(b)` would be fun too (or dot(c))

MichaelClerx:
4. The MMT docs will need updating (they are under `docs` somewhere). Happy to help with this

myokit/_model_api.py
martinjrobins: done

myokit/_err.py
martinjrobins: added

myokit/_model_api.py
```diff
@@ -3236,11 +3243,17 @@ def add_equation(eq_list, done, eq):
         return out
 
     def state(self):
```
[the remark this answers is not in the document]
martinjrobins: done

myokit/_sim/cable.py
```diff
@@ -176,7 +176,7 @@ def __init__(self, model, protocol=None, ncells=50, rl=False):
                 ' next')
 
         # Set state and default state
-        self._state = self._model.state() * ncells
+        self._state = self._model.state_values() * ncells
```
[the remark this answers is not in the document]
martinjrobins: done

myokit/tests/test_model_building.py
```diff
@@ -133,6 +133,22 @@ def test_model_creation(self):
...
+        # non-constant expression initial value should error
+        b.promote()
+        x.demote()
+        with self.assertRaises(myokit.NonConstantExpressionError):
+            x.promote(myokit.Name(b))
+        b.demote()
+
+        # set constant expression initial value
+        x.promote(myokit.Name(b1))
+        self.assertEqual(x.state_value(), 1)
```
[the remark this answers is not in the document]
martinjrobins: done

myokit/tests/test_parsing.py
```diff
@@ -1389,7 +1389,21 @@ def test_initial_values(self):
             """
         myokit.parse(code)
 
-        # Non-literal value
+        # initial expression
+        code = """
+            [[model]]
+            c.q = 10 * 2 + c.p
```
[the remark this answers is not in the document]
martinjrobins: done

myokit/tests/test_parsing.py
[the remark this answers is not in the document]
martinjrobins: do you expect this to parse ok? if I have:
```
[[model]]
c.p = dot(c.p)

[engine]
time = 0 bind time

[c]
dot(p) = 1
```

then `is_constant()` evaluates false for `dot(c.p)`...., what is the expected behaviour here?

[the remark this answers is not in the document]
MichaelClerx: Ha!
Hadn't thought about that. I'm quite happy for it to be False because, superficially, dp/dt depends on p and t. Maybe open a ticket to investigate if that should be True in the future :D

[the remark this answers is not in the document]
MichaelClerx: I _think_ I'm fine with myokit thinking dx/dt is never constant (even if the expression for it is).
It already thinks that `y/y` is an expression that depends on y, for example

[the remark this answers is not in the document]
MichaelClerx: Yes. I'm sure it thinks `x = x - x` is a cyclical dependency too!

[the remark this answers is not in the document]
martinjrobins: complete now I think

myokit/tests/test_simulation_cvodes.py
```diff
@@ -936,6 +936,28 @@ def test_derivatives(self):
         # Compare
         self.assertLess(np.max(np.abs(e)), 0.1)
 
+    def test_initial_state_expression(self):
```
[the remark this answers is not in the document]
martinjrobins: what do you mean here? the `default_value` will be unchanged (just 'p'), rather than updated

myokit/_sim/cvodessim.py
```diff
@@ -510,7 +510,7 @@ def reset(self):
 
         """
         self._time = 0
-        self._state = list(self._default_state)
+        self._state = [float(expr) for expr in self._default_state]
```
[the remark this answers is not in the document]
martinjrobins: I realised that there was a bug in `set_default_state`, in that it evaluates whatever was given by the user and sets the default state to floats, rather than expr. Presumably we want to keep it as expr.

The problem was the `Model.map_to_state` function, which evaluates the state given. Do we want to keep this behavior and create another `map_to_state_expr` that preserves expressions, or update `map_to_state` so that it doesn't evaluate?

martinjrobins:
> 1. Should initial values be allowed depend on child variables? The usual rule is that you can only access your own child vars, not anybody elses. Not sure if I feel like the init cond. belong to the variable or to the whole model. What's it like in your use cases?

The init cond is placed outside the compartment, and thus far away from the variable defn, so atm I think it belongs to the whole model, and therefore can't depend on child variables

myokit/_sim/cvodessim.py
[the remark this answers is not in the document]
MichaelClerx: Yikes!
It's kinda making me rethink the default stat idea. But it's really necessary for your use case that the simulation can have a modifiable constant thats used as/by an initial state, no?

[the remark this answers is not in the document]
MichaelClerx: How about set_default_state tries to parse it as a sequence of numbers or expression, and if it fails it hands off to map_to_state?
... but then you'd want to re-use that so OK let's have a map_to_state_expr

martinjrobins:
> 2. Can we get a test where we promote so that `init(x) = a; a = b; b = 1 + a`? This should be ignored by promote(), but state_values should raise an error (e.g. by first calling validate()) instead of getting caught in some endless loop

We currently can't promote in this case, because a.is_constant() returns `False`, is this the expected behaviour?

MichaelClerx:
> We currently can't promote in this case, because a.is_constant() returns False, is this the expected behaviour?

Yeah that makes sense. Does mean we need a test in `validate` that checks that it's still constant after manipulations?

MichaelClerx:
> > 1. Should initial values be allowed depend on child variables? The usual rule is that you can only access your own child vars, not anybody elses. Not sure if I feel like the init cond. belong to the variable or to the whole model. What's it like in your use cases?
> 
> The init cond is placed outside the compartment, and thus far away from the variable defn, so atm I think it belongs to the whole model, and therefore can't depend on child variables

Let's make sure we mention this in the docs for promote() and in the mmt syntax docs

myokit/tests/test_simulation_cvodes.py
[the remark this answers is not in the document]
MichaelClerx: Just trying to make sure there's tests for using set_default_state with expressions somewhere! (doesn't have to be here)

martinjrobins:
> > 2. Can we get a test where we promote so that `init(x) = a; a = b; b = 1 + a`? This should be ignored by promote(), but state_values should raise an error (e.g. by first calling validate()) instead of getting caught in some endless loop
> 
> We currently can't promote in this case, because a.is_constant() returns `False`, is this the expected behaviour?

sadness :( So you can promote `y`, *then* setup the cyclic dependence, and then call `state_values()` which then gets trapped in the cyclic evaluation..... Do we want to check that model is valid (using `Model.validate()`) in `state_values()` before we evaluate?

MichaelClerx:
We'll need a bunch of tests for the partial derivatives sim. E.g.

- Basic getting partial derivative dy/dp where y(t=0) = f(p)
- Does it work correctly if we update p with simulation.set_constant
- Does it work correctly if we remove the p dep with set_state() or by running a sim (should lose to p dependency)
- Can we run a sim in two parts? So integrate from t=0 to t=1 with y(t=0)=p. At this point the p-dep should be gone, but the internal representation of dy/dp will still be non-zero so that's fine. Now we sim from t=1 to t=2. Should get the same result for dy/dp as doing a single sim from t=0 to t=2.
- Similarly, but now change p halfway through the sim. Shouldn't affect state, or dy/dp.

MichaelClerx:
> sadness :( So you can promote `y`, _then_ setup the cyclic dependence, and then call `state_values()` which then gets trapped in the cyclic evaluation..... Do we want to check that model is valid (using `Model.validate()`) in `state_values()` before we evaluate?

Yeah I'd be happy with a validate() call before state_values. Validation is expensive but cached.

martinjrobins:
yay, all done I think, time for a re-review!

CHANGELOG.md
```diff
@@ -4,9 +4,15 @@ This page lists the main changes made to Myokit in each release.
...
 - Added
+  - [#898](https://github.com/myokit/myokit/pull/899) allow initial conditions for state variables to be expressions (previously only literals allowed).
+  - [#898](https://github.com/myokit/myokit/pull/899) added `myokit.NonConstantExpressionError` for an error where a constant expression is expected.
+  - [#898](https://github.com/myokit/myokit/pull/899) `Model.state_values` returns a list of floats, the result of evaluating each element of `Model.state`.
 - Changed
+  - [#898](https://github.com/myokit/myokit/pull/899) `Model.state` now returns a list of expressions.
+  - [#898](https://github.com/myokit/myokit/pull/899) `Variable.promote` can take an expression, a number, or a string (parsed as an expression) for the arg `state_value`.
 - Deprecated
 - Removed
+  - [#898](https://github.com/myokit/myokit/pull/899) removed `myokit.NonLiteralValueError` as no longer used.
```
MichaelClerx, line 15:
```suggestion
  - [#898](https://github.com/myokit/myokit/pull/899) Initial conditions for state variables can now be (constant-valued) expressions. Previously, only literals were allowed.
  - [#898](https://github.com/myokit/myokit/pull/899) Added `myokit.NonConstantExpressionError` for an error where a constant expression is expected.
  - [#898](https://github.com/myokit/myokit/pull/899) `Model.state_values` now returns a list of floats, the result of evaluating each element of `Model.state`.
- Changed
  - [#898](https://github.com/myokit/myokit/pull/899) `Model.state` now returns a list of expressions.
  - [#898](https://github.com/myokit/myokit/pull/899) `Variable.promote` can now take an expression, a number, or a string (parsed as an expression) for the argument `state_value`.
- Deprecated
- Removed
  - [#898](https://github.com/myokit/myokit/pull/899) `NonLiteralValueError` is now no longer raised by Myokit and so has been removed.
```

MichaelClerx:
@martinjrobins do you think we should tackle https://github.com/myokit/myokit/issues/435 at the same time? (Since we're already making breaking changes to the state-related API...)

MichaelClerx (approved):
Nice work @martinjrobins ! I'll go over a bit more, and add in #435 too, but then hopefully merge tomorrow (thursday)

martinjrobins:
> Nice work @martinjrobins ! I'll go over a bit more, and add in #435 too, but then hopefully merge tomorrow (thursday)

Great, thanks @MichaelClerx, good idea to add in #435, I didn't see this one.

MichaelClerx:
@martinjrobins I'm still not 100% convinced about the simulations using expressions?

For example:
```
s = myokit.Simulation(m, p)
s.run(100)  # starts from evaluated expressions
s.run(100)  # starts from floats. OK!
s.reset()
s.pre(100)   # starts from evaluated expressions
s.run(100)   # starts from floats
```

a bit complicated for the users perhaps?

What is the use-case for having it start from expressions: can you give a real world example?

martinjrobins:
The usecase is running multiple simulation with different parameters, say that an initial state value depended on a constant `c.p`, when you do this:

```python
s = myokit.Simulation(m)
s.run(1)
s.set_constant('c.p', 2)
s.reset()
s.run(1)
```

unless you keep the expressions, the initial state value won't be updated and therefore the behaviour is unexpected from the user perspective

MichaelClerx:
> The usecase is running multiple simulation with different parameters, say that an initial state value depended on a constant `c.p`, when you do this:
> 
> ```python
> s = myokit.Simulation(m)
> s.run(1)
> s.set_constant('c.p', 2)
> s.reset()
> s.run(1)
> ```
> 
> unless you keep the expressions, the initial state value won't be updated and therefore the behaviour is unexpected from the user perspective
> 
> realworld example is doing an optimisation or inference where parameters are constantly changing

Would you have other variables depending on `c.p` ? Or is it an easier way of writing `x = s.state(); x[cp.indice()] = 2; s.set_state(x)` ?

MichaelClerx:
The bit that worries me is that you can now do:
```
s.pre(10)
s.set_constant('c.p', 2)
```
and it will no longer update the initial value.

Also the logic for calculating dX/dp now gets a little bit different, as `dX/d init(z)` might no longer be allowed if `init(z)` is no longer a literal

martinjrobins:
but `s.pre` should override the initial value coming from the model right, so this is expected behaviour?

Yes, you are correct that `dX/d init(z)` is no longer valid as init(z) is no longer a parameter of the model. I would argue that the initial values of state variables are by default *not* parameters of the model, unless specified by the user, i.e. `init(z) = p`, then `dXdp`

martinjrobins:
> Would you have other variables depending on c.p ? Or is it an easier way of writing x = s.state(); x[cp.indice()] = 2; s.set_state(x) ?

Yes, you could have other variables depending on `c.p`. If that dependency was in the rhs equations, you would expect the result of your sim to be different after you call `set_constant`. Same thing if the dependency was in the initial conditions.

But calling `set_state` has a different meaning altogether, `set_state` manually sets the current state of the simulation, this ignoring the initial conditions from the model (you would need to call `reset` to reset the current state to the initial conditions coming from the model)

MichaelClerx:
Oh dear, initial values can have units now

MichaelClerx:
So how would unit checking initial conditions work?

```
membrane.V = -80    # Historically OK
membrane.V = -80 [mV]     # Should now be allowed already
membrane.V = -80 [V]        # Should now be  a unit error
membrane.V = -0.08 [V] * 1000 [mV/V]     # Show now be OK
membrane.V = -80 + offset    # where offset = -20 mV
membrane.V = -80 * scaling    # where scaling = 1.2
```

martinjrobins:
I'm not sure how units are handled in myokit yet. what are the rules for units for `dot(membrane.V)`? Should be the same for the initial conditions right, except the final unit will be different?

martinjrobins:
just call `eval_unit` on the rhs, if it doesn't match the lhs then error? if rhs has no unit then assume its the same unit as the lhs?

MichaelClerx:
> I'm not sure how units are handled in myokit yet. what are the rules for units for `dot(membrane.V)`? Should be the same for the initial conditions right, except the final unit will be different?

Writing
```
time = 0 [s]
    in [s]
dot(x) = 1 [m/s]
```
defines a variable `x` with units of `m`

So if you use `x` anywhere in an equation it's in `m`, while `dot(x)` _in an equation_ is in `m/s`: `y = x + 3[m]` is fine, as is `z = dot(x) + 2 [m/s]`

martinjrobins:
Sure that makes sense. But how do you do the unit checking for the rhs of `dot(x) = rhs`? I'm saying that the same logic should apply for the examples you gave earlier

MichaelClerx:
For everything except the initial conditions, Myokit's unit checking follows one of two rule sets:

1. Strict unit check: Everything that doesn't declare a unit is dimensionless. So `x = 3 in [m]` is wrong (declared unit is `m`, but rhs is `3` which gets interpreted as `3 [1]` i.e. 3 dimensionless
2. Tolerant unit check: Undeclared units are assumed to be ok, i.e. `x = 3 in [m]` gets OK'd, and things like `2 + 1[m]` are interpreted as `2 [m] + ` [m]`

For initial values there isn't any unit checking whatsoever - in part because there wasn't even a mechanism to give units to the initial values! So not sure what to do with them

martinjrobins:
for backwards compatibility, sounds like 2. Tolerant unit check is the way to go for initial conditions?

MichaelClerx:
Yeah I agree. The thing is, users choose which kind of unit checking they want (strict, tolerant, or don't care). The problem is that to stay backwards compatible this will introduce some kind of limbo state where you say strict-but-tolerant-about-initial-values. Unless we want people to start writing units for all initial value expressions...
So this PR is raising quite a few issues --- just for the convenience of being able to change initial values with `set_constant`. 
On the other side, it's making me think more about models and things I hadn't really considered, i.e.
- There are 3 values/references associated with a state var: derivative, current value, and initial value. At the moment, Myokit mixes current and initial. I.e. `membrane.V = -80` but then `dot(V) = f(V)`. Pedantically, this is wrong! It should be `init(membrane.V) = -80`, `dot(V) = f(V)`, `V = defined by solver`
- Apparently I forgot about units for initial states
- At some point there was an idea that a model would have a "current" state

martinjrobins:
I don't think this is a convenience PR, its more changing the space of models that the mmt language can express. Mathematically you currently define your ODE something like this:

dot(y) = f(y, t, p)
y(0) = g()

where p are your constant variables. but this PR now allows you to define your ODE like this:

dot(y) = f(y, t, p)
y(0) = g(p)

So my thinking about units is that they are closely related to types in traditional programming languages. So allowing users to determie the level of unit checking is like a language allowing you to choose if you want strict or dynamic/duck typing. i.e. a nightmare for the person writing the compiler ;) 

You could get around this by defining a null unit, if a factor in an expression has null unit, then it can be promoted to an actual unit if it is used in a binary operation with another factor that has a unit. So null unit is compatible with all units.

MichaelClerx:
> You could get around this by defining a null unit, if a factor in an expression has null unit, then it can be promoted to an actual unit if it is used in a binary operation with another factor that has a unit. So null unit is compatible with all units.

😄 😄 😄  That is _exactly_ how the tolerant unit checking works!

martinjrobins:
So for the strict unit checking you can't use null units? Except for initial conditions?

MichaelClerx:
Another issue: Myokit tracks references to variables, and this is cached. At the moment there's 4 lists in each var: refs to, refs by, refs to state value, refs by state eqs. Would have to add another 2 and check all the caching etc. everywhere.

@martinjrobins I'm getting quite worried about this PR:

cons:
- few more days of work, at least
- big api changes
- more complicated system of state/default_state for users to get to grips with
- potential for new bugs in model api, in all simulation code, in all export code
- awkward new rules for unit checking of initial values
pros:
- can support cellml initial values w vars
- can write `for var, value in zip(vars, parameters): sim.set_constant(var, value)` instead of slightly more complex forms like
```
for var, value in zip(param_vars, parameters[:n_param_vars]):
    sim.set_constant(var, value)
state = sim.state()
for var, value in zip(init_vars, parameters[n_param_vars:]):
    state[var.indice()] = value
sim.set_state(state)
```

MichaelClerx:
And even `for var, value in zip(vars, parameters): sim.set_constant(var, value)` is slightly hacky because it means you need to have written your model so that one of the initial conditions is set as a variable introduced only for this purpose. If that's the only use case we're after, how about just adding a convenience method like `set_parameter(variable, value)` that can set both constants and states?

MichaelClerx:
> So for the strict unit checking you can't use null units? Except for initial conditions?

No in strict mode the lack of a unit (or a null unit) is treated as a statement that this variable or expression is unitless.

So now the rules would have to become:

- Tolerant: no declared unit is treated as compatible with anything
- Strict: no declared unit is treated as dimensionless, except for initial conditions where the tolerant rules apply

martinjrobins:
> Strict: no declared unit is treated as dimensionless, except for initial conditions that are numbers, in which case no unit needs to be given

Yea, agree with this it seems to map onto the current type checking well

I would add another "pro" arguement as well (this is the main reason why I made the PR):
- can write an mmt model with an initial expression dependent on a constant variable

tbh, I don't really care about simulating that model, as you say you can do anything in Python. For me its all about expressing the initial expression in the mmt file so it can be shared with others, stored in  a database etc.

MichaelClerx:
> tbh, I don't really care about simulating that model, as you say you can do anything in Python. For me its all about expressing the initial expression in the mmt file so it can be shared with others, stored in a database etc.

So in light of this, and all the confusion that we could get from keeping expressions in a simulation's default state, I've decided that the simulations simply use the evaluated `model.initial_values(as_floats=True)`

MichaelClerx:
By an amazing coincident our phd student frankie has just arrived at a use case for knowing dy/dp where p is used by inits, so status is back to "I need to think about it" 😆

MichaelClerx:
I'm working on this this week, and a little bit in the next few weeks if needed.
Have split it into a full-blown project to simplify and make things closable

PR #907, opened 2022-10-26
Simplified Model.evaluate_derivatives

### LINKED ISSUE #906: Rewrite `evaluate_derivatives(state=...)` to use `subst` instead of changing the model state
Current method changes the model state, then changes it back again (without a try-finally block)

MichaelClerx:
Simplified `Model.evaluate_derivatives`, making it use `subst()` instead of changing any variable RHS equations. 

Closes #906.

### DIFF
--- a/myokit/_expressions.py
+++ b/myokit/_expressions.py
@@ -389,10 +389,7 @@ class Expression(object):
 
     def eval(self, subst=None, precision=myokit.DOUBLE_PRECISION):
         """
-        Evaluates this expression and returns the result. This operation will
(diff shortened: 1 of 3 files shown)


MichaelClerx:
@martinjrobins this avoids any changes to the model in `evaluate_derivatives`.

(That method is older than e.g. simulations in Myokit, so code was a bit ropey, apologies for that!)

martinjrobins (approved):
this is great and helps a lot, thanks @MichaelClerx

PR #909, opened 2022-11-07
Updated workflows to test Python 3.11 for workgroups

MichaelClerx:
- Add 3.11 testing on linux, windows, mac
- Drop 3.6 testing on linux (already dropped on others)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,17 +1,17 @@
 [flake8]
 filename=
     *.py,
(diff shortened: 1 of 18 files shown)


MichaelClerx:
Current issue: default opencl device doesn't have double precision support, so that bit isn't tested. Does it select the fpga emulator by default?

PR #918, opened 2022-11-24
i435 rename "state value" to "initial value"

MichaelClerx:
See #435 

Doing this now, as it removes backward incompatibilities in #899 

**Model API**
- [x] Change API
- [x] It should now be an error to delete a variable that is referenced in an initial value. Add and make tests.
- [x] Leave unit checking for separate ticket #938

**Tests**
- [x] Test model parsing with expressions (always in global context)
- [x] ~Test model parsing can raise `NonConstantValueError`~
- [x] Test `Model.initial_values` (floats or expr)
- [x] Test `Model.set_initial_values` (takes lots of input args, if string then parsed w global context, ~can raise `NonConstantValueError`~)
- [x] Test `Model.code` shows expressions
- [x] Test `Model.clone`
- [x] Test `Model.format_state()` and `Model.format_state_derivatives` with expr (get evaluated).
- [x] Test `Model.import_component` with init expr
- [x] Test `Model.show_evaluation_of`
- [x] Test `Model.validate()` raises `NonConstantExpressionError`
- [x] Test `Variable.initial_value` (float or expr)
- [x] Test `Variable.set_initial_value` (takes lots of input args, if string then parsed w global context, ~can raise `NonConstantValueError`~)
- [x] Test `Variable.promote` with new possible input args (via `set_initial_value`?).
- [x] Test `ModelComparison` with expr inits.
- [x] Test deprecation warning from `Model.state` and date in comment in the code
- [x] Test deprecation warning from `Model.set_state` and date in comment in the code
- [x] Test deprecation warning from `Model.inits` and date in comment in the code
- [x] Test deprecation warning from `Variable.state_value` and date in comment in the code
- [x] Test deprecation warning from `Variable.set_state_value` and date in comment in the code
- [x] Test deprecation warnings from `Model.save_state` and `Model.load_state`
- [x] Test deprecation warnings from promote keyword arg state_value
- [x] Add tests for self-ref in initial values (`c.x = c.x`, `c.x = 1 + c.x`)
- [x] Add tests for `InitialValue` and `PartialDerivative` in init if needed
- [x] Add test for conditionals in init if needed
- [x] Drop `NonConstantExpressionError`


MichaelClerx:
@martinjrobins I've tried to write myself a big to-do list in the form of `INIT_CHANGES.md`. I'll re-use some of this in an [example notebook](https://github.com/myokit/myokit-examples) at some point, but also could be a good basis for extended unit tests.
Can you have a look via [Files changed](https://github.com/myokit/myokit/pull/918/files) or [here](https://github.com/myokit/myokit/blob/e5a75d07fb4c1b25cbafa9269ce109e8b005135d/INIT_CHANGES.md) and tell me what you think?

INIT_CHANGES.md
```diff
@@ -0,0 +1,347 @@
...
+print('This is OK, because c.p is literal:')
+sim = myokit.Simulation(model, sensitivities=(['c.y'], ['c.p']))
+
+print('This is not OK, because init(c.y) is not literal:')
+sim = myokit.Simulation(model, sensitivities=(['c.y'], ['init(c.y)']))
+```
+
+### SimulationOpenCL and FiberTissueSimulation
+
+Not sure!
```
martinjrobins, line 168: these should behave in a similar way, I tried to keep all simulations consistant

INIT_CHANGES.md
```diff
@@ -0,0 +1,347 @@
...
+
+- `JacobianTracer`
+- `JacobianCalculator`
+- `RhsBenchmarker`
+
+### Legacy sims
+
+Should probably just remove these now!
+
+- `LegacySimulation`
```
martinjrobins, line 194: agreed

MichaelClerx:
Making some progress... hopefully next week!

INIT_CHANGES.md
MichaelClerx: done

MichaelClerx:
Trying to write out the full scope of the changes below, with an eye to reuse in release notes

# Changes for users

In Myokit versions before 1.34, a model's initial values were represented internally as numbers (although it was allowed to specify them using a numbers-only expression like 'membrane.V = -80 + 1`).
This allowed you to specify systems of ODEs as:
```
dy/dt = f(y, t, p)
y(t=t0) = y0
```
where `y` is the state, `t` is the time variable, and `p` is the set of all time-invariant variables (i.e. everything that isn't a state, a bound variable, or something that depends on a state or bound variable).
The value `t0` is an unspecified time when the state equals the initial conditions.

Starting from version 1.34, a model's initial values are stored as expressions, which are allowed to refer to constant-valued variables:
```
dy/dt = f(y, t, p)
y(t=0) = g(p)
```

Note, however, that this is only implemented at the model level: Simulations will still treat the state as a literal constant.

## Model API

- The `Model` methods `state()` and `set_state()` are deprecated in favour of `initial_values` and `set_initial_values` which handle both floats and expressions. Instead of `Model.inits()` you can now use `Model.initial_values(as_equations=True)`.
- The `Variable` methods `state_value()` and `set_state_value()` are now deprecated in favour of `initial_value()` and `set_initial_value()`, which can handle both floats and expressions.
- The method to create states, `Variable.promote()`, now handles expressions.

Note that these methods all operate on the initial values that a model specify, making the name change appropriate.
Model methods `format_state` and `map_to_state` all operate on _any_ model state, so retain the current name.

A full list of changes is given in the [changelog](https://github.com/myokit/myokit/blob/main/CHANGELOG.md).

## Unit checking

Before version 1.34 initial values' units were never checked.
Starting from version 1.34, initial value expressions that reference variables will be unit checked.

```
component.a = 1                 # Not unit checked before or after 1.34
component.b = 1 + 1             # Not unit checked before or after 1.34
component.c = component.x       # Unit checked in 1.34, invalid before
component.d = 1 / component.x   # Unit checked in 1.34, invalid before
```

Easy way to implement: Initial values that have literal values and don't specify a unit automatically get the unit from the variable.

## Simulations

When a simulation is created, the model's initial state is evaluated and used as the simulation's state and default state.

Because the initial values are evaluated upon creation, using `set_constant` on a parameter referenced by an initial value does not change the simulation state or default state.

Similarly, when calculating the sensitivity of the solution with respect to a parameter that's referenced by an initial value, the sensitivity through the initial value will _not_ be taken into account.

MichaelClerx:
Some more updates:

- const check should only happen in `validate()` (myokit lets you make invalid models in lots of ways, but then validates before you do anything with the model e.g. simulate or export. The reason is that the API should be flexible, i.e. create a variable, give it a value, move variables around, etc. all without having to leave a valid model after every operation.
- NonLiteralValueError was only around because parsing needed to be able to distinguish it from other IntegrityErrors. With new parsing set up (promote with dummy first, then later give real value) this is no longer required so NonConst error can be dropped

Issue #881, opened 2022-07-05
Use opencl runtime from package manager?


MichaelClerx:
Intel now seems to recommend it:

- https://www.intel.com/content/www/us/en/developer/articles/tool/opencl-drivers.html#inpage-nav-undefined-1-undefined
- https://github.com/intel/compute-runtime/blob/master/DISTRIBUTIONS.md

MichaelClerx:
BUT, this is for GPUs, not for CPUs, so no use on github actions

MichaelClerx:
There is now a thing called the "oneapi" blah blah that contains the cpu runtime for opencl

Here it says you need it: https://www.intel.com/content/www/us/en/developer/articles/technical/intel-cpu-runtime-for-opencl-applications-with-sycl-support.html

And here it explains how to add the non-standard repo to apt: https://www.intel.com/content/www/us/en/develop/documentation/installation-guide-for-intel-oneapi-toolkits-linux/top/installation/install-using-package-managers/apt.html

PR #923, opened 2022-12-14
Can we install opencl on ubuntu purely with apt-get?

MichaelClerx:
See #881

Suddenly became urgent as old ubuntu opencl unit tests no longer seem to work


MichaelClerx:
No, the package installs a GPU, not a CPU driver

MichaelClerx:
But this bit is specifically about CPUs: https://www.intel.com/content/www/us/en/developer/articles/tool/opencl-drivers.html#proc-graph-section

Issue #925, opened 2022-12-15
Speed up unit tests with caching


MichaelClerx:
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

- [x] Enable caching of pip dependencies
- [ ] Cache the apt-stuff 
- [ ] Similar for windows
- [ ] Similar for mac

MichaelClerx:
Apart from the pip cache, it doesn't seem to work

PR #926, opened 2022-12-15
Testing with pip cache

MichaelClerx:
Part of  #925


MichaelClerx:
Doesn't speed things up. The cached packages are already cached somewhere else it seems! (They are all very common)

PR #936, opened 2023-02-11
Frankie edit

frankiepe:
Added methods:


myokit/_sim/cvodessim.py
```diff
@@ -361,6 +361,37 @@ def default_state_sensitivities(self):
             return [list(x) for x in self._s_default_state]
         return None
 
+    def state_sensitivities(self):
```
MichaelClerx, line 364: Minor first point: **going forward**, I'm trying to keep all methods sorted alphabetically. It's a very silly thing, but

1. The "natural" order stops making sense if you haven't looked at the code for a while, and isn't the same for all people, so for lack of a better idea alphabetical is now often the norm
2. Existing myokit code often isn't alphabetical, but moving the methods around messes with the git history (makes it look like massive changes happened), so that's why I'm saying "going forward". 
3. Sorry

myokit/_sim/cvodessim.py
```diff
@@ -361,6 +361,37 @@ def default_state_sensitivities(self):
...
+    def set_state_sensitivities(self, s_values):
+        """
+        Sets the sensitivities with respect to state variables if
+        sensitivities are enabled
+
+        The N x M list of lists 's_values', defines the sensitivities
+        dz/dx for the N state variables and M independent variables
+        """
+        if self._sensitivities:
+            self._s_state = s_values
```
MichaelClerx, line 382: This is potentially dangerous: because `s_values` is a list or array or some "non-atomic" type, it can be modified by anyone who has a reference. So we might make some decisions based on its contents, and then the user can still modify it, and our internals get messed up! So copy before storing.

myokit/_sim/cvodessim.py
```diff
@@ -361,6 +361,37 @@ def default_state_sensitivities(self):
...
+
+    def set_state_sensitivities(self, s_values):
+        """
+        Sets the sensitivities with respect to state variables if
+        sensitivities are enabled
+
+        The N x M list of lists 's_values', defines the sensitivities
+        dz/dx for the N state variables and M independent variables
+        """
+        if self._sensitivities:
```
MichaelClerx, line 381: Probably wants an error if sensitivities are not enabled (because user seems to think they are: API should be helpful here and complain)

myokit/_sim/cvodessim.py
MichaelClerx, line 382: Wants some basic checks before accepting? E.g. is it the right size/shape or even data type?

myokit/_sim/cvodessim.py
```diff
@@ -361,6 +361,37 @@ def default_state_sensitivities(self):
...
+            self._s_state = s_values
+
+    def set_default_sensitivities(self, s_values):
```
MichaelClerx, line 384: default_state_sensitivities, for consistency?

myokit/_sim/cvodessim.py
```diff
@@ -361,6 +361,37 @@ def default_state_sensitivities(self):
...
+            return [list(x) for x in self._s_state]
+        return None
+
+    def set_state_sensitivities(self, s_values):
+        """
+        Sets the sensitivities with respect to state variables if
+        sensitivities are enabled
+
+        The N x M list of lists 's_values', defines the sensitivities
+        dz/dx for the N state variables and M independent variables
```
MichaelClerx, line 379: Can use ``dz/dx`` to get a code formatting, if desired.

Also needs a second reference to both z and x to say which is which

myokit/tests/test_simulation_cvodes.py
```diff
@@ -311,17 +314,35 @@ def test_default_state_sensitivites(self):
...
+        sim.set_state_sensitivities(s_values)
+        sim.set_default_sensitivities(s_values)
+        sd = sim.default_state_sensitivities()
+        s = sim.state_sensitivities()
+        self.assertEqual(len(sd), 2)
+        self.assertEqual(len(s), 2)
+        self.assertEqual(sd[0][0], 2)
+        self.assertEqual(s[0][0], 2)
+        self.assertEqual(sd[1][0], 3)
+        self.assertEqual(s[1][0], 3)
```
MichaelClerx, line 344: Nice!

myokit/_sim/cvodessim.py
frankiepe: What do you mean by this "Can use dz/dx to get a code formatting, if desired."?

MichaelClerx: Sorry, I had added backticks (`), but github ate them up!

myokit/_sim/cvodessim.py
```diff
@@ -361,6 +362,69 @@ def default_state_sensitivities(self):
...
+        Here we define z as a vector of dependent variables (state or
+        state-derived) of length N, and x as a vector of independent
+        variables (parameters or initial conditions) of length M.
+        """
+        if self._sensitivities:
+            if isinstance(s_values, list):
+                if [len(s_values), len(s_values[0])] == [len(
+                        self._s_default_state), len(self._s_default_state[0])]:
+                    s_val_copy = copy.deepcopy(s_values)
+                    self._s_default_state = s_val_copy
```
MichaelClerx, line 417: Might be easier here to cast to a numpy array and check it's shape?
Or even something like `self._s_default_state = [list(x) for x in s_values]` to avoid needing `copy` just for this line

Issue #320, opened 2019-05-13
Value of `pace` when running without a protocol


MichaelClerx:
- Simulation
  - [x] Set to 0 for all pacing labels defined by user (also after use!)
  - [x] No protocol set? Then still register `pace` to stay backwards compatible.
  - [x] But if user explicitly says `protocol={}`, then don't set pace to 0!
- [x] OpenCLSimulation: Set `pace` to 0 explicitly (also after use!)
- [x] Fiber tissue: Set `pace` to 0 explicitly (also after use!)
- [x] Cable: Set `pace` to 0 (also after use!)
- [x] MM & HH: These only allow Vm to be paced, so no changes necessary

PR #947, opened 2023-03-21
I534-multiple-protocols

### LINKED ISSUE #534: Multiple protocols
There's some reasons it might be good to have multiple protocols:

- Clamping both membrane potential and a concentration to some time signal (e.g. Vm and Ca)

This would require quite major changes in Myokit.

~One way to implement this might be to remove the "binding" construct, and replace it with some special function (`myokit.Expression`), e.g.~

```
I_stim = input('pace') * stim_amplitude
I_diff = input('diffusion_current')
```
~where 'diffusion_current' would be a name set by the simulation (but customisable?), while the name 'pace' would be set by the user, e.g. by having a `name: pace` field in the protocol definition, or a constructor like `p = myokit.Protocol('pace')`.~

A different way to implement it would be to set it in the simulation, e.g.

```
s.set_protocol('pace', p)
s.set_protocol('other_label', q)
```

martinjrobins:
Fixes #534

1. `myokit.Simulation` constructor takes a dict mapping binding labels to protocols (which can be None)
1. `myokit.Simulation.set_protocol` can take an optional label (default `pace`) to set the protocol for that binding label. The protocol can be an event-based protocol or a fixed form protocol
1. new python class `myokit.FixedProtocol` to encapsulate a fixed form protocol (previously just a tuple of times/values)


myokit/_sim/cmodel.py
```diff
@@ -137,6 +139,45 @@ def __init__(self, model, sensitivities):
         self.parameters = parameters
         self.parameter_derived = parameter_derived
 
+    def _prepare_bindings(self, model, labels):
```
MichaelClerx, line 142: I was imagining this either getting inlined (and possibly simplified?) anywhere we used it, or becoming a private method somewhere. Happy with this for now, but definitely part of the bigger project to get rid of or hide this somehow

myokit/_sim/cvodessim.c
```diff
@@ -195,11 +195,19 @@ Model model;        /* A model object */
...
+union PSys {
+    ESys event;
+    FSys fixed;
+};
+enum PSysType {
+    EVENT,
+    FIXED
+};
+union PSys *pacing_systems;   /* array of pacing system (event or fixed) */
+enum PSysType *pacing_types; /* array of pacing system types */
```
MichaelClerx, line 207: Not sure I understand the syntax of repeating the union and enum keywords here...

myokit/_sim/cvodessim.py
```diff
@@ -121,9 +121,13 @@ class Simulation(myokit.CModule):
...
-    ``protocol``
-        An optional :class:`myokit.Protocol` to use as input for variables
-        bound to ``pace``.
+    ``protocols``
+        An optional dict mapping paced variable labels (str) to a
+        :class:`myokit.Protocol` or None. if None, the paced variable will be
+        set to 0, or the pacing protocol can be set later using
+        :meth:set_protocol. For backwards compatibility, a single protocol
+        of type :class:`myokit.Protocol` can be passed in instead of a dict,
+        the label in this case will be 'pace'.
```
MichaelClerx, line 130: I imagine the single or no protocol cases will remain more popular than the multi-protocol sims, so probably better to phrase as e.g. "A protocol, a dict mapping ..., or None. The protocol can be..."

myokit/_sim/cvodessim.py
```diff
@@ -951,66 +974,45 @@ def set_min_step_size(self, dtmin=None):
 
     def set_fixed_form_protocol(self, times=None, values=None):
```
MichaelClerx, line 975: I suppose we'd eventually want to move to some structure where we have a `Protocol` (?) base class, with subclasses `StepProtocol` and `TimeSeriesProtocol` (or similar) ?
Probably a good idea to delay that until after this PR?

MichaelClerx (commented):
Hey @martinjrobins ,
It looks great!
Some questions here and there, but this looks simpler than I was expecting and very nicely done. Wish I could hire you :D

myokit/_fixed_protocol.py
```diff
@@ -0,0 +1,103 @@
...
+# See http://myokit.org for copyright, sharing, and licensing details.
+#
+from __future__ import absolute_import, division
+from __future__ import print_function, unicode_literals
+from __future__ import annotations
+from bisect import bisect_left
+from typing import List
+
+
+class FixedProtocol(object):
```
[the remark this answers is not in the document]
martinjrobins: Sure, that sounds good.
 I was hoping to extend to different interpolation methods in the future (my particular usecase is piecewise constant)

myokit/_fixed_protocol.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        return self._times
+
+    def values(self) -> List[float]:
+        """
+        Returns a list of the values in this protocol.
+        """
+        return self._values
+
+    def pace(self, t: float) -> float:
```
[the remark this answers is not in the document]
martinjrobins: yup, will do

myokit/_fixed_protocol.py
```diff
@@ -0,0 +1,103 @@
...
+        self._values = list(values_tpl)
+        if method is None:
+            method = 'linear'
+        else:
+            method = str(method).lower()
+            if method not in ['linear']:
+                raise ValueError('Unknown interpolation method: ' + method)
+        self._method = method
+
+    def times(self) -> List[float]:
```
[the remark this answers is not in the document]
martinjrobins: ok, I'll remove the typing (its very handy tho, would encourage slow adoption!)

myokit/_fixed_protocol.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def pace(self, t: float) -> float:
+        """
+        Returns the value of the pacing variable at time ``t``.
+        """
+        if t < self._times[0]:
+            return self._values[0]
+        if t > self._times[-1]:
+            return self._values[-1]
+        i = bisect_left(self._times, t) - 1
```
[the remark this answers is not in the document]
martinjrobins: will change to `bisect_right` as per issue

myokit/_sim/cvodessim.py
martinjrobins: yup, will do

myokit/_sim/cvodessim.py
```diff
@@ -158,13 +162,19 @@ def __init__(self, model, protocol=None, sensitivities=None, path=None):
...
 
         # Set protocol
-        self._protocol = None
-        self._fixed_form_protocol = None
-        self.set_protocol(protocol)
-        del protocol
+        if isinstance(protocols, myokit.Protocol):
+            protocols = {'pace': protocols}
+        if protocols is None:
+            protocols = {'pace': myokit.Protocol()}
```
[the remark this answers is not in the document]
martinjrobins: mainly for backwards compatability I think. Previously you were always binding the `pace` variable, so we still want to do that?

myokit/_sim/cvodessim.py
martinjrobins: yea, I'd agree with that (and the delaying ;) )

myokit/_sim/cvodessim.py
```diff
@@ -951,66 +974,45 @@ def set_min_step_size(self, dtmin=None):
...
+
+        if label not in self._pacing_labels:
+            raise ValueError('Unknown pacing label: ' + str(label))
+
+        index = self._pacing_labels.index(label)
 
         # Set new protocol
         if protocol is None:
-            self._protocol = None
+            self._protocols[index] = myokit.Protocol()
```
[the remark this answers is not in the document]
martinjrobins: the length of the list can't be updated, as this corresponds to the bound variables, which can't change. As the previous comment, you need to at least bind the `pace` variable for backwards compat

myokit/_sim/cvodessim.py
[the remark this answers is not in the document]
martinjrobins: ok, I'll give that a go

myokit/_fixed_protocol.py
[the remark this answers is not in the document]
martinjrobins: agree that value would make more sense, but I think we should leave it as pace for now to match `Protocol` and perhaps change it after

myokit/_sim/cvodessim.py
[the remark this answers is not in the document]
martinjrobins: nope, this has some unfortunate effects. Previously the `pace` binding was always kept, but if you don't bind it here, then it is removed which would break users code subsequently. Think we have to keep as is until #320 is resolved

martinjrobins:
@MichaelClerx : there are errors on mac + windows that I'm having trouble with. I tried my macbook but the problem doesn't show up there, do you have the same error on your machine? The CI isn't giving any information about the failure...

martinjrobins:
also, I'm not sure about that doc test failure. Is there a particular way you want the docs arranged? I put the `TimeSeriesProtocol` in the `Protocol.rst` file, but your test doesn't seem to like that....

myokit/_sim/cvodessim.c
martinjrobins: the joy of C.....

MichaelClerx:
Rewrote a doctstring that was generating a warning and also the intel dependencies magically started working again :D

myokit/_sim/pacing.h
```diff
@@ -750,21 +754,31 @@ FSys_Destroy(FSys sys)
...
-    if(!PyList_Check(values_list)) return FSys_POPULATE_INVALID_VALUES;
+    if (protocol == Py_None) return FSys_POPULATE_INVALID_PROTOCOL;
+
+    // Get PyList from protocol (will need to decref!)
+    // Cast to (char*) happens because CallMethod accepts a mutable char*
+    // This should have been const char* and has been fixed in python 3
+    PyObject* times_list = PyObject_CallMethod(protocol, (char*)"times", NULL); // Returns a new reference
+    if(times_list == NULL) return FSys_POPULATE_INVALID_PROTOCOL;
+    if(!PyList_Check(times_list)) {
+        Py_DECREF(times_list);
```
MichaelClerx, line 773: Do `times_list` and `values_list` need to be decreffed on every possible exit of this function after this? Looks like a dangling reference otherwise?

myokit/tests/test_simulation_cvodes.py
```diff
@@ -91,6 +134,13 @@ def test_no_protocol(self):
...
         self.assertTrue(np.all(d['engine.pace'] == 0.0))
 
+    def test_wrong_label_set_pacing(self):
+        # Test set_pacing with incorrect label
+        self.sim.reset()
+        self.sim.pre(50)
+        with self.assertRaisesRegex(ValueError, 'Unknown pacing label'):
+            self.sim.set_protocol(None, label='does not exist')
+
     def test_fixed_form_protocol(self):
```
MichaelClerx, line 144: This needs to be updated to 

1. Use the new TimeSeriesProtocol type calls everywhere
2. Have a single call using the old syntax wrapped in a WarningCollector that checks that the deprecation warning is shown

MichaelClerx (changes requested):
Looks amazing Martin. Two small to-dos then it's going in. I think you've also convinced me that it's useful to maintain bindings as a separate concept to labels

myokit/_sim/pacing.h
MichaelClerx: Fixed it!

myokit/tests/test_simulation_cvodes.py
MichaelClerx: done

MichaelClerx (approved):

PR #969, opened 2023-06-07
Added mention of initial time y(t0)=y0 in mmt syntax doc.

### LINKED ISSUE #928: initial condition for time
So in a myokit model you can set initial conditions for the state variables. But how can you set the value for time where these initial conditions are defined? Is this possible or is t:=0 for the initial conditions?

MichaelClerx:
Closes #928


martinjrobins (approved):
looks great 👍

PR #979, opened 2023-06-07
Add PyQt6 and PySide6 support

MichaelClerx:
- [x] IDE
- [x] DataLogViewer
- [x] DataBlockViewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@ This page lists the main changes made to Myokit in each release.
 
 ## Unreleased
 - Added
+  - [#979](https://github.com/myokit/myokit/pull/979) Added support for PyQt6 and PySide6.
 - Changed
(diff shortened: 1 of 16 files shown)


MichaelClerx:
Probably also a good time to drop PyQt4 and PySide support.

[Wikipedia](https://en.wikipedia.org/wiki/Qt_version_history#Qt_4) says 5 was released in 2012, and support for 4 ended in 2015

MichaelClerx:
*PyQt6 uses `exec` as a function name, which causes a syntax error in Python 2*
So unless we want to have some file that renames it exec_ that we import from only on PySide6, this might have to wait until we drop Python2 ?

MichaelClerx:
> _PyQt6 uses `exec` as a function name, which causes a syntax error in Python 2_ So unless we want to have some file that renames it exec_ that we import from only on PySide6, this might have to wait until we drop Python2 ?

Done 😁

Issue #36, opened 2018-02-23
ABF reader scaling wrong in some files


MichaelClerx:
The scaling factor in the voltage signal (second channel, AD1) of `c260315_0000.abf` (model-cell-3) is off by a factor ten. It works properly in clampfit.

MichaelClerx:
Strange. The test case is a v1.65 file, so should be able to see exact code in support pack.

```
//==============================================================================================
// FUNCTION:   GetADCtoUUFactors
// PURPOSE:    Calculates the scaling factors used to convert ADC values to UserUnits.
// PARAMETERS:
//    nChannel        - The physical channel number to get the factors for.
//    pfADCToUUFactor - Pointers to return locations for scale and offset.
//    pfADCToUUShift    UserUnits = ADCValue * fADCToUUFactor + fADCToUUShift;
//
void WINAPI ABFH_GetADCtoUUFactors( const ABFFileHeader *pFH, int nChannel, 
                                    float *pfADCToUUFactor, float *pfADCToUUShift )
{
   ABFH_ASSERT(pFH);
   WPTRASSERT(pfADCToUUFactor);
   WPTRASSERT(pfADCToUUShift);
   ASSERT(nChannel < ABF_ADCCOUNT);

   float fTotalScaleFactor = pFH->fInstrumentScaleFactor[nChannel] *
                             pFH->fADCProgrammableGain[nChannel];
   if (pFH->nSignalType != 0)
      fTotalScaleFactor *= pFH->fSignalGain[nChannel];

   // Adjust for the telegraphed gain.
   if( pFH->nTelegraphEnable[nChannel] )
      fTotalScaleFactor *= pFH->fTelegraphAdditGain[nChannel];

   ASSERT(fTotalScaleFactor != 0.0F);
   if (fTotalScaleFactor==0.0F)
      fTotalScaleFactor = 1.0F;

   // InputRange and InputOffset is the range and offset of the signal in
   // user units when it hits the Analog-to-Digital converter

   float fInputRange = pFH->fADCRange / fTotalScaleFactor;
   float fInputOffset= -pFH->fInstrumentOffset[nChannel];
   if (pFH->nSignalType != 0)
      fInputOffset += pFH->fSignalOffset[nChannel];

   *pfADCToUUFactor = fInputRange / pFH->lADCResolution;
   *pfADCToUUShift  = -fInputOffset;
}
```

apart from rounding etc that's the same as what the current code does

MichaelClerx:
[...]

Issue #984, opened 2023-06-21
Improve conversion to data log in ephys formats


MichaelClerx:
- ~ABF: Replace different methods with a single one with switches~
- Make sweep joining optional
- Use 0. etc naming for sweeps when not joined
- Use channel names in log
- Similar for ATF
- Similar for WCP

Full to-do in #985

MichaelClerx:
Grepping for datalog vs data_log shows very little public API using either:

```
./myokit/tests/test_datablock.py:    def test_from_data_log(self):
./myokit/tests/test_formats_wcp.py:    def test_data_log_conversion(self):
[michael@localhost myokit]$ grep "datalog" ./myokit/ -Ir
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer.ico',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-16.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-24.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-32.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-48.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-64.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-96.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-128.xpm',
./myokit/gui/datalog_viewer.py:#        'icon-datalog-viewer-256.xpm',
./myokit/gui/datalog_viewer.py:            '.csv': self.load_datalog,
./myokit/gui/datalog_viewer.py:            '.zip': self.load_datalog,
./myokit/gui/datalog_viewer.py:    def load_datalog(self, filename):
./myokit/_bin/install-lin/myokit-datalog-viewer.desktop:#Icon=<?=icons?>/icon-datalog-viewer.png
./myokit/tests/test_datalog.py:        path = os.path.join(DIR_IO, 'datalog.csv')
./myokit/tests/test_datalog.py:        path = os.path.join(DIR_IO, 'datalog-1-empty.csv')
./myokit/tests/test_datalog.py:        path = os.path.join(DIR_IO, 'datalog-2-windows.csv')
./myokit/tests/test_datalog.py:        path = os.path.join(DIR_IO, 'datalog-3-old-mac.csv')
[...]

PR #998, opened 2023-07-11
Updates to the SweepSource class, more ABF class rewrites

MichaelClerx:
See #984


MichaelClerx:
Now also fixes https://github.com/myokit/myokit/issues/36

PR #1018, opened 2023-08-07
Patchmaster updates to read more amplifier properties

MichaelClerx:
@chonlei

This adds some code to deal with the `.amp` segment of a file, and updates the viewer to show incomplete traces (it wasn't showing any for Alex's data, presumably because it couldn't parse the stimulus and so couldn't determine that the traces were from completed recordings).

The code isn't finished.
In particularly, it sticks a load of information in a property called `._temp`.

If you can figure out what these fields (or any others you want to add) mean, what units they're in etc., then we can add proper methods for them (with docstrings that say what they mean).

This script reads a file and prints all the *unparsed* amplifier info (so not the stuff where I actually know exactly what it means):
```
#!/usr/bin/env python3
import myokit
import myokit.formats.heka as heka


files = [
    '220502_008.dat',
]

for fname in files:
    with heka.PatchMasterFile(fname) as f:
        for group in f:
            group_label = group.label()

            print(f'Splitting {group_label} ({fname})')
            count = {}

            for series in group:
                print(f'Series label: {series.label()}')
                for i, a in enumerate(series.amplifier_states()):
                    print(f'Amplifier {i}')
                    for k, v in a._temp.items():
                        print(f'  {k}: {v}')
```

### DIFF
--- a/myokit/formats/heka/_patchmaster.py
+++ b/myokit/formats/heka/_patchmaster.py
@@ -1682,7 +1682,7 @@ class AmplifierState:
     def c_slow_auto_settings(self):
         """
         Returns a tuple with the settings used when automatically estimating
-        C-slow: (Pulse amplitude in mV, number of cycles, Timeouet in s).
+        C-slow: (Pulse amplitude in mV, number of cycles, Timeout in s).
 
         See :meth:`c_slow`.
         """
--- a/myokit/formats/heka/_patchmaster.py
+++ b/myokit/formats/heka/_patchmaster.py
@@ -1016,7 +1016,7 @@ class Series(TreeNode, myokit.formats.SweepSource):
         # Resistance, capacitance, etc.
         a = self.amplifier_state()
         out.append('Information from amplifier state:')
-        out.append('  Current gain: {a.current_gain()} mV/pA')
+        out.append(f'  Current gain: {a.current_gain()} mV/pA')
         if a.ljp():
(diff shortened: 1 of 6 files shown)


chonlei:
I tried the script. Does it print out only those that you pre-set or is it printing everything from the `.amp` already?

MichaelClerx:
The script above prints only the properties added to `._temp`: https://github.com/myokit/myokit/blob/30ea6a191fbd21a7ca8e41b7d27001494dcac841/myokit/formats/heka/_patchmaster.py#L1554

chonlei:
@MichaelClerx I added a two more to the existing list that I thought would be really useful to have. And the temp list looks good to me so far, though I re-ordered it so that tau_rs and Imon1bandwidth go with their groups---hopefully will be easier later when rewriting them into proper methods. Hope that's OK.

myokit/formats/heka/_patchmaster.py
```diff
@@ -1571,12 +1679,28 @@ def c_slow(self):
         """
         return self._cs * 1e12
 
+    def c_slow_auto_settings(self):
+        """
+        Returns a tuple with the settings used when automatically estimating
+        C-slow: (Pulse amplitude in mV, number of cycles, Timeouet in s).
```
MichaelClerx, line 1685:
```suggestion
        C-slow: (Pulse amplitude in mV, number of cycles, Timeout in s).
```

PR #1055, opened 2024-03-15
Fix bug with PrefixPlus expressions like `+(a - b) / c`

### LINKED ISSUE #1054: Bug with PrefixPlus operator in code export and Simulation
```
#!/usr/bin/env python
import myokit

x = myokit.parse_expression('(b - c) / d')
y = myokit.parse_expression('+(b - c) / d')
e = myokit.formats.ewriter('python')
print(e.ex(x))
print(e.ex(y))
```
results in
```
(str:b - str:c) / str:d
str:b - str:c / str:d
```

(To be clear, it's only when you use the _prefix_ plus operator, so "+1 / 2" that it comes into play. "3 + 4" works as expected)

MichaelClerx:
Fixes #1054 

An issue was discovered in the way Myokit exported expressions involving a `PrefixPlus` (the unary plus operator) to `C` , `OpenCL`, and other languages. As the C and OpenCL export was used in simulations, this can potentially affect simulation output. This pull request fixes this and a similar issue (see below)

## What was going wrong?

Myokit's export to C and several other languages rendered expressions like `+(a - b) / c` as `a - b / c`.

This resulted from the early (2013!) decision to ignore the `+` when generating C code, as it didn't change the meaning of the expression (`+5` means the same as `5`). Unfortunately, in bypassing the `+` printing code, it also bypassed the bracket-adding code. This is not a problem for expressions that don't need brackets (`+5` to `5`, `+exp(V)` to `exp(V)`), but affects the meaning of "infix" operators, like addition and multiplication.

### What was not going wrong

- Plus and Minus in general worked as expected: `(a + b) / c` was rendered as `(a + b) / c`.
- The PrefixMinus worked as expected: `-(a - b) / c` was rendered as `-(a - b) / c`

### How was it fixed?

This pull request makes the expression writers always output the + and the brackets. Leaving the removal of potentially unneeded + operators to the compiler. It also adds many more unit tests for Myokit's expression conversion system, involving ones that check the output numerically for Python, Numpy, and Ansi C.

### What else was going wrong?

In adding these tests, a second issue was discovered in the export to Python, NumPy, and Stan. These languages use a "right-associative" power operator, that is, they interpret the expression `a^b^c` (or `a**b**c` in Python syntax) to mean `a^(b^c)`. Languages like C however, interpret `a^b^c` as `(a^b)^c`. Myokit now enforces the second (left-associative) behaviour by adding brackets to double-power operations in Python, Numpy, and Stan.

### What is the impact of these bugs?

As it affected the numerical output, it could have lead to silent mistakes in simulation results. However, the use of a prefix plus in front of an infix expression is quite uncommon. Similarly, the use of a power of a power is not something you would expect in a model.

A script is provided below that will check a model to see if it contains a potential issue.


MichaelClerx:
This script will either scan a single model or recursively scan a directory, looking for potential issues in the equations:

```
#!/usr/bin/env python3
#
# This script scans Myokit models for equations involving a 'PrefixPlus', that
# is a unary plus operation, which are affected by the bug described at
# https://github.com/myokit/myokit/pull/1055
#
import os
import myokit

# Set this to the file or directory you want to check
path = '../../mmt-curated' #akwaboah-2021.mmt'


def has_issue(expression):
    """ Scan an expression for potential issues. """
    if isinstance(expression, myokit.PrefixPlus):
        if isinstance(expression[0], myokit.InfixExpression):
            return True
    if isinstance(expression, myokit.Power):
        if isinstance(expression[0], myokit.Power):
            return True
    for op in expression:
        if has_issue(op):
            return True
    return False


def scan_var_owner(owner, issues):
    """ Scan a component or variable. """
    for var in owner:
        if has_issue(var.rhs()):
            issues.append((f'Equation', var, var.rhs()))
        if var.is_state() and has_issue(var.initial_value()):
            issues.append((f'Initial value', var, var.initial_value()))
        scan_var_owner(var, issues)


def scan_model(filename):
    print(f'Checking model {filename}...', end='')
    model = myokit.load_model(filename)

    issues = []
    for component in model:
        scan_var_owner(component, issues)
    if issues:
        print(' [potential issue detected]')
        for where, var, expression in issues:
            print(f'{where} for {var.qname()}: {expression.code()}')
    else:
        print(' [ok]')


def scan_dir(path):
    for leaf in os.listdir(path):
        full = os.path.join(path, leaf)
        if leaf in ('.', '..'):
            continue
        if os.path.isdir(full):
            scan_dir(full)
        elif leaf.endswith('.mmt'):
            scan_model(full)


if __name__ == '__main__':
    if os.path.isdir(path):
        scan_dir(path)
    else:
        scan_model(path)
```

Example output on an affected model:
```
Checking model akwaboah-2021.mmt... [potential issue detected]
Equation for calcium.Ca_i: if(mode.cversion == 1, cell.Cm, 1) * (+(calcium.jCa_diff * cell.V_sub - calcium.jup * cell.V_up) / cell.Vi - (calcium.CM_tot * dot(cabuf.cmi) + calcium.TC_tot * dot(cabuf.tc) + calcium.TMC_tot * dot(cabuf.tmc)))
```

Example output on a model directory:
```
Checking model ./sampson-2010.mmt... [ok]
Checking model ./paci-2013-ventricular-vs.mmt... [ok]
Checking model ./paci-2013-ventricular.mmt... [ok]
Checking model ./livshitz-2007.mmt... [ok]
Checking model ./noble-1962.mmt... [ok]
Checking model ./ni-2017.mmt... [ok]
Checking model ./koivumaki-2011.mmt... [ok]
```

Issue #1059, opened 2024-04-24
Robustness to irrelevant overflows for AnalyticalSimulations


mjowen:
During the update step for the HH Analytical Simulation (the dynamically defined HHmodel._function), an overflow during the timescale/tau calculation leads to nan outputs in the current and the states for `_t=0`, even though at `_t=0` the states are already known.

Below is an example of a HHmodel._function which can cause this issue (I've simplified it from a larger example where I found this bug but it should still be a viable _function)
``` python
import numpy

def _f(_y0, _t, V, p1, p2, p3):
    _y = [0]*1
    pace = -80.0
    ikur_ui_inf = p1
    ikur_ui_tau = abs(1.0 / (1 + numpy.exp((V + p2) / p3)))
    EK = -70
    _y[0] = ui = ikur_ui_inf + (_y0[0] - ikur_ui_inf) * numpy.exp(-_t / ikur_ui_tau)
    _i = ui * (V - EK)
    return _y, _i

_f(numpy.array([0.5]), numpy.array([0,1]), -80, 0.75, -100, -0.05)
```

If an overflow is triggered in the numpy.exp for the `ikur_ui_tau` statement, then `ikur_ui_tau=0.0`
During the state update, `numpy.exp(-_t / ikur_ui_tau)` returns a nan if `_t==0` and 0 if `t>0` (returns inf for `t<0` which is why the nan is here).
This nan is then returned in the current and states from myokit and this will happen at every `t=0` which means every change in voltage.

Since we already know the state at `_t=0` we could include a line in _f to set _y to _y0, ie.
``` python
    _y[0] = ui = ikur_ui_inf + (_y0[0] - ikur_ui_inf) * numpy.exp(-_t / ikur_ui_tau)
    _y[0][_t==0] = ui[_t==0] = _y0[0]
```

MichaelClerx:
Thanks @mjowen !

> During the state update, numpy.exp(-_t / ikur_ui_tau) returns a nan if _t==0 and 0 if t>0 (returns inf for t<0 which is why the nan is here).

I don't quite understand. If `ikur_ui_tau == 0` it should return `y[0] = [nan, inf, inf, inf...]`, right? So `inf` for `t>0` rather than `0`?

mirams:
That's what it does return, but it could return `y[0] = [initial condition, inf, inf, inf...]` which would let an objective function be calculated as Inf (and used as a big number), rather than NaN (not usable!).

mjowen:
[...]

Issue #943, opened 2023-02-22
Rewrite HH Model class


MichaelClerx:
- No link to myokit.Model once it's created
- API to create from scratch
- see #808

Issue #571, opened 2020-06-17
Compiled HH solver?


MichaelClerx:
Might not be too hard.
Could even do a GPU one?

MichaelClerx:
Alternatively an llvm/numba based on?
If the accellerated code is super simple this could be the fastest way?

Issue #1048, opened 2024-03-04
Support cvodes 7


MichaelClerx:
https://github.com/LLNL/sundials/releases

MichaelClerx:
Closed in #1063

PR #1060, opened 2024-04-30
1059 overflow

mjowen:
Fixed issue #1059 by forcing initial condition when _t=0 in the HH analytical simulation
Added a test to ensure this fix doesn't break with the changes in #943 and #571 
The test verifies that:

- An overflow warning is triggered
- No nans are present in the current or the overflown state variable
- The state is correctly solved for tau=0 (starts at initial state then jumps to steady state)

### DIFF
--- a/myokit/tests/test_lib_hh.py
+++ b/myokit/tests/test_lib_hh.py
@@ -15,6 +15,8 @@ import myokit.lib.hh as hh
 
 from myokit.tests import DIR_DATA
 
+from myokit.tests import WarningCollector
+
 
 MODEL = """
 [[model]]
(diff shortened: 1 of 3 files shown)


myokit/tests/test_lib_hh.py
```diff
@@ -914,6 +916,39 @@ def test_against_cvode_v_independent(self):
         e = np.abs(d1['binding.I'] - d2['binding.I'])
         self.assertLess(np.max(e), 2e-4)
 
+    def test_tau_overflow(self):
+        # Overflows leading to tau=0 should still report current
```
MichaelClerx, line 920:
```suggestion
        # Overflows leading to tau=0 should still report current
        # https://github.com/myokit/myokit/issues/1059
```

MichaelClerx (dismissed):
Looks good, one minor tweak would be to mention the github issue in the test
Thanks!

mjowen:
By the way, I did check Gary's suggestion of doing indexing on _t!=0 on the first call for each state to avoid the extra np.exp call but the extra lines needed to preallocate the variables and the extra indexing meant it didn't save any time

MichaelClerx:
Afraid merge will have to wait for #1048 and #1062

MichaelClerx (approved):

MichaelClerx:
Thanks @mjowen !
Will do a release soon so you get this behaviour in the pip installed version

PR #1061, opened 2024-04-30
Bumping minimum supported Python to 3.8

MichaelClerx:
To fix Mac test

### DIFF
--- a/.github/workflows/unit-tests-macos.yml
+++ b/.github/workflows/unit-tests-macos.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         # Selected python versions only
-        python-version: ['3.7', '3.11']
+        python-version: ['3.8', '3.11']
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -169,7 +169,7 @@ Because most libraries do, Myokit uses American spelling (e.g. "color", "fiber",
 
 ## Python and C versions
 
-Myokit is developed for Python 3.7 and newer, and tested on whatever GitHub Actions (easily) allows (3.7+ at time of writing).
+Myokit is developed for Python 3.8 and newer, and tested on whatever GitHub Actions (easily) allows (3.7+ at time of writing).
```
MichaelClerx, line 172:
```suggestion
Myokit is developed for Python 3.8 and newer, and tested on whatever GitHub Actions (easily) allows (3.8+ at time of writing).
```

MichaelClerx:
We're now testing on 3.8 and 3.11, so this bit is fixed.

MichaelClerx:
However, there's a new issue with MacOS tests throwing an OpenCL error, where they really should not

PR #1088, opened 2024-10-16
Updates to heka meta data reading

MichaelClerx:
- Logviewer now shows meta data for myokit DataLogs
- Fixed fast capacitance value read by heka patchmaster format
- Added more meta data to heka reader, e.g. filter 1 and filter 2 settings


MichaelClerx:
Hi @chonlei ! Don't really need a review, but might be good for you to know we can get some filter 1 and filter 2 info now!

MichaelClerx:
Would still like to add meta on the stimulus filter too: 

Screenshot from 2024-10-16 19-59-08

MichaelClerx:
The `C_fast_tau` value is wrong too. Most likely because the true c fast filtering is a two parter. No idea how to get from this value to the full one, so might have to just omit?

Issue #1090, opened 2024-10-18
DataLog / DataLogViewer: Allow reading CSV files with duplicate column headers


MichaelClerx:
Nothing stopping people from doing this in a CSV header row...

PR #1092, opened 2024-10-18
Added test for reading duplicate keys from CSV

MichaelClerx:
No rules against this...
See #1090 

With this update, it turns "x", "x" into "x-1", "x-2".

It turns "x", "x", "x-2" into "x-1", "x-3", "x-2"


MichaelClerx:
Closes #1090

PR #1098, opened 2024-10-31
feat: add SBML export

### LINKED ISSUE #1091: export to SBML
It would be nice to have an export to SBML, in addition to the existing importer

comment from @MichaelClerx:

> it wouldn't be too hard to add SBML export. There is already an intermediate description (objects representing an SBML model). So you'd need to write (A) something that turns a myokit.Model into a myokit.formats.SBML.model, and (B) something to write SBML models as XML. Neither should be an issue, & I'm happy to help if anyone wants to take it on

martinjrobins:
Fixes #1091


MichaelClerx:
Thanks @martinjrobins !
Do you think it'd make sense to move the xml writing code to a separate class, like we have for CellML? https://github.com/myokit/myokit/blob/main/myokit/formats/cellml/v1/_writer.py

martinjrobins:
> Thanks @martinjrobins ! Do you think it'd make sense to move the xml writing code to a separate class, like we have for CellML? https://github.com/myokit/myokit/blob/main/myokit/formats/cellml/v1/_writer.py

its totally up to you :) . I found this easier because the `to_xml` functions are highly correlated to the internal attributes of each class (`Model`, `Species` etc). So its nice to have these side by side when you are writing the code (or later on when you might want to add a new attribute to a class). But happy to refactor this to a separate class so its more consistent with the rest of myokit

MichaelClerx:
That'd be good if you could. I think I found it useful in delineating between the structure and the representation! Otherwise little bits of overlap kept creeping in :D

MichaelClerx:
@DavAug how's things? Would you be interested / have time to do a review for this?

DavAug:
Hello 😊
Wow nice piece of work @martinjrobins ! 

No problem, I'll have a look @MichaelClerx . It just might take me a couple days to properly look through it.

myokit/tests/test_formats_sbml.py
```diff
@@ -15,7 +15,63 @@
...
+    def test_capability_reporting(self):
+        # Test if the right capabilities are reported.
+        e = myokit.formats.exporter('sbml')
+        self.assertTrue(e.supports_model())
+        self.assertFalse(e.supports_runnable())
+
+    def test_stimulus_generation(self):
+        # Tests if protocols allow a stimulus current to be added
+
+        e = myokit.formats.exporter('cellml')
```
martinjrobins, line 35: whoops, will turn this on!

martinjrobins: @DavAug. The exporter uses rules to set the rhs equations for all species (no reactions). However, when I then use your parser to read in the exported file it fails at this line: https://github.com/myokit/myokit/blob/403a029a8ff05d1a847bbf14070905750c7bbafc/myokit/formats/sbml/_parser.py#L543

The SBML spec indicates that it is allowed to use rules for non boundary species, as long as you don't use reactions as well. But the code here seems to indicate that the parser will fail if rules are used for a non-boundary species?

DavAug: Hi @martinjrobins ,
I have to admit, it has been a long time since we implemented the parser. So if in doubt, I would trust you more than me! Can you specify for me what "boundary species" are? 

Just looking at the comments in the code (see below), we may have made this restriction because it's not quite obvious how you would combine assignment rules with reactions (assignments being in some some sense absolute statements about a variable's derivatives RHS, while reactions can be composed to define the RHS.

But it's very possible that we made a mistake here.

```
# Check that this assignable can be changed with assignment rules
            if isinstance(var, myokit.formats.sbml.Species):
                if not var.is_boundary():
                    raise SBMLParsingError(
                        'Assignment or rate rule set for species that is'
                        ' created or destroyed in a reaction.', element)
            # TODO: Could also gather 'constant' attributes for all, and check
            # here. But maybe we're not all that interested in validation...
```

martinjrobins: not really sure what boundary species are, but from the name I guess they live on the boundary and so the volume does not affect them. In any case, the SBML spec is that a boundary species can be involved in rules, or reactions, but not both (so they deal with the combination or rules + reactions by forbiding this case)

DavAug: Hi @martinjrobins and @MichaelClerx ,
I had quick refresher on the SBML specification (specifically page 51 and 52 in file:///Users/DAUGU/Downloads/sbml.level-3.version-2.core.release-2.pdf).

In particular, have a look at the last row in the table on page 52, where it says that when `constant=False` and `boundaryCondition=False`, the RHS is given by "reactions or rules (but not both), and events".

We do not support events in Myokit.

So that is why we have the above check that when we have an assignment rule to a species that is not a boundary species, we throw an error because otherwise the RHS of the species would be set by BOTH the assignment rule and the reaction.

If the species is a boundary species, the reaction does not influence the RHS of the species and the RHS is set by an assignment rule, see for example page 128 in file:///Users/DAUGU/Downloads/sbml.level-3.version-2.core.release-2.pdf

Do you agree with that understanding of the SBML documentation?

DavAug: Alternatively, (if we don't have a good way to permitting assignment rules for non boundary species) maybe it is worth trying to parse individual terms of the RHS change rates into reactions, but it would involve some more processing.

MichaelClerx: Hey both,

I'm not sure about when a species can/should be set with an assignment rule, but I'm surprised we're talking about species at all, to be honest 😅 

The main example model we looked at for import was the HH model, which only uses "parameters": https://github.com/myokit/myokit/blob/main/myokit/tests/data/formats/sbml/HodgkinHuxley.xml
So I'd expected our output to be in terms of parameters as well.

A stronger argument is that species are defined as:

> A species in SBML refers to a pool of entities that (a) are considered indistinguishable from each other for the purposes of the model, (b) may participate in reactions, and (c) are located in a specific compartment

Section 4.6, page 49. Myokit doesn't have a concept of "reactions" or of "compartments", so whether b and c hold can't be determined from the equations alone. So the default I think, should be to export everything as a "parameter".

If we do want species to appear in the exported models, then we should think of (and document) a way to get this information into the model. For example, by doing something like:
```
dot(x) = 3
    sbml_role: species
```

MichaelClerx: (fwiw, I think martin might be right, but have made a separate issue #1101)

DavAug: I agree. I think SBML's functionality of being able to define compartments and reactions is quite useful when you manually define large systems of ODEs, but when we automatically export myokit models, perhaps it's fine to just stick to parameters and assignment rules.

martinjrobins: > @martinjrobins Do you have any input/expected-output files that you're basing this on, or is it more of an "any working export will do"? Cause in that case I'd definitely ignore the complexity of species & reactions :D
> 
> (Of course, it does mean that sbml --> myokit --> sbml will make parameters out of all your species, and replace all the reactions with explicit equations. But I think that would be extremely difficult to avoid.)

All these are the inputs: https://github.com/pkpdapp-team/pkpdapp/tree/master/pkpdapp/pkpdapp/migrations/models
Outputs are just SBML files that, when simulated, give identical results the the mmt files.

I think any state variable in a myokit model should become an SBML species right? As I understand it, a paramter can't have a rate equation?

MichaelClerx: > The use of the term parameter in SBML sometimes leads to confusion among readers who have a particular
notion of what something called “parameter” should be. It has been the source of heated debate, but despite this, no one has yet found an adequate replacement term that does not have different connotations to different people and hence leads to confusion among some subset of users. Perhaps it would have been better to have two constructs, one called “constants” and the other called “variables”. The current approach in SBML is simply more parsimonious, using a single Parameter construct with the boolean flag constant to indicate which flavor the parameter is. In any case, readers are implored to look past their particular definition of a “parameter” and simply view SBML’s Parameter as a single mechanism for defining both constants and (additional) variables in a model. (We write additional because the species in a model are usually considered to be the central variables.)

MichaelClerx: This is Lucian's (an sbml co-author) take on it:

> Yeah, while that is technically legal, I think I would flag it as a boundary species anyway.  Or, you know, actually use 'parameter'!
>
> While the numbers in the simulation will be correct regardless, the produced model will have increased flexibility in the future if you use the correct types--when you do things like calculating the Jacobian or even just the stoichiometry matrix, having superfluous species will just confuse things.  It will also help when annotating.

So definitely parameters then

martinjrobins: coming back to this now. So is the plan to export all myokit variables as parameters then? What did you want me to do about compartments then, since a parameter does not have a compartment. Each parameter could just have a name based on their qname rather than their name? and otherwise we ignore compartments...?

MichaelClerx: > So is the plan to export all myokit variables as parameters then?
Yup!

> What did you want me to do about compartments then, since a parameter does not have a compartment. 
Ignore them, I think

> Each parameter could just have a name based on their qname rather than their name? and otherwise we ignore compartments...?
Yeah it's a shame there doesn't seem to be a "conceptual" grouping mechanism, just the physical "species-compartment" one.

martinjrobins: although I guess this will break the expressions for species which are in compartments :)

martinjrobins: ahh, I guess I can just use `set_lhs_function` to set a custom name

MichaelClerx: Sorry I was cooking :D 
Yeah that should work!

docs/source/api_index/myokit.formats.rst
```diff
@@ -211,11 +211,13 @@ myokit.formats.sbml
 - :class:`myokit.formats.sbml.Model`
 - :class:`myokit.formats.sbml.ModifierSpeciesReference`
 - :meth:`myokit.formats.sbml.importers`
+- :meth:`myokit.formats.sbml.exporters`
```
MichaelClerx, line 214:
```suggestion
- :meth:`myokit.formats.sbml.exporters`
- :meth:`myokit.formats.sbml.importers`
- :class:`myokit.formats.sbml.Model`
- :class:`myokit.formats.sbml.ModifierSpeciesReference`
```

myokit/formats/sbml/_exporter.py
```diff
@@ -0,0 +1,54 @@
+#
+# Exports to SBML.
+#
+# This file is part of Myokit.
+# See http://myokit.org for copyright, sharing, and licensing details.
+#
+import myokit
+from myokit.formats.sbml._writer import write_file
```
MichaelClerx, line 8: Can `write_file` be imported into `myokit.formats.sbml` so that it's all possible publicly as well? (the exporter is always a bit of a shortcut)

myokit/formats/sbml/_writer.py
```diff
@@ -0,0 +1,341 @@
...
+def write_file(path: str, model: Model):
+    """
+    Writes a CellML 1.0 or 1.1 model to the given path.
+    """
+    return SBMLWriter.write_file(path, model)
+
+
+def write_string(model: Model) -> str:
+    """
+    Writes a CellML 1.0 or 1.1 model to a string and returns it.
```
MichaelClerx, line 30:
```suggestion
    Writes an SBML model to a string and returns it.
```

myokit/formats/sbml/__init__.py
```diff
@@ -33,5 +34,18 @@ def importers():
     """
     return dict(_importers)
```
MichaelClerx, line 36: Can we get a 

```
from ._writer import (  # noqa
    write_file,
    write_string,
    SBMLWriter,
)
```

here? (With updated imports in other files, and mention of these 3 in the docs, sorry!)

martinjrobins: sure, done

myokit/formats/sbml/_exporter.py
martinjrobins: yup, done

myokit/formats/sbml/_writer.py
```diff
@@ -0,0 +1,339 @@
+#
+# CellML 1.0/1.1 Writer: Writes a CellML Model to disk
```
MichaelClerx, line 2:
```suggestion
# SBML Writer: Writes an SBML Model to disk
```

myokit/formats/sbml/_writer.py
```diff
@@ -0,0 +1,339 @@
...
+        if parameter.is_constant():
+            parameter_xml.attrib['constant'] = 'true'
+            value = parameter.value().eval()
+            parameter_xml.attrib['value'] = str(value)
+            return parameter_xml, None, None
+        else:
+            initial_assignment, rule = SBMLWriter._quantity(parameter)
+            return parameter_xml, initial_assignment, rule
+
+    def _math(expression: myokit.Expression) -> etree.Element:
```
MichaelClerx, line 120: I'm guessing this is all pretty fast, so there's no point making this class a single-use object that can reuse the same mathmlexpressionwriter ?

myokit/formats/sbml/_writer.py
```diff
@@ -0,0 +1,339 @@
...
+        root = etree.Element(
+            'sbml',
+            xmlns='http://www.sbml.org/sbml/level3/version2/core',
+            level='3',
+            version='2'
+        )
+
+        # setup a map from unit to string
+        unit_map_to_str = {
+            unit: string for string, unit in Model._base_units.items()
```
MichaelClerx, line 257: Some non-private way of getting these? Only if easy

myokit/formats/sbml/_writer.py
```diff
@@ -0,0 +1,339 @@
...
+        if model.extent_units() != myokit.units.dimensionless:
+            model_root.attrib['extentUnits'] = unit_map_to_str[
+                model.extent_units()
+            ]
+        if model.volume_units() != myokit.units.dimensionless:
+            model_root.attrib['volumeUnits'] = unit_map_to_str[
+                model.volume_units()
+            ]
+
+        if model.units():
```
MichaelClerx, line 289: Should we have a has_units() etc? Since this copies a whole list?

myokit/tests/test_sbml_export.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def test_time_unit(self):
+        # Test setting the time unit
+        model = Model()
+        model.set_time_units(myokit.units.second)
+        sbml_str = write_string(model).decode("utf8")
+        self.assertIn('timeUnits="second"', sbml_str)
+
+        model = Model()
+        model.set_time_units(1e3 * myokit.units.second)
```
MichaelClerx, line 40: Who needs hours when we've got kiloseconds! Gonna see if I can get a watch with these

MichaelClerx (approved):
Looks great! 
Thanks for this Martin.
Few small questions

myokit/formats/sbml/_api.py
```diff
@@ -247,6 +250,93 @@ def __init__(self, name=None):
...
+        time = model.time()
+
+        # Method to obtain or infer variable unit
+        def variable_unit(variable, time_unit):
+            """Returns variable.unit(), or attempts to infer it if not set."""
+            unit = variable.unit()
+            if unit is not None:
+                return unit
+
+            rhs = variable.rhs()
```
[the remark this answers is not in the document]
martinjrobins: ok, I'll leave it. It wasn't my approach, I took it from the cellml exporter

myokit/formats/sbml/_api.py
```diff
@@ -247,6 +250,93 @@ def __init__(self, name=None):
...
+            if rhs is not None:
+                try:
+                    # Tolerant evaluation, see above. Result may be None.
```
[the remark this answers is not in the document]
martinjrobins: I expanded on this comment a bit

myokit/formats/sbml/_api.py
```diff
@@ -247,6 +250,93 @@ def __init__(self, name=None):
...
+
+        # Create unames in model for nested variables
+        model.create_unique_names()
+
+        def add_variable(variable: myokit.Variable):
+            if variable is time:
+                return
+            v = m.add_parameter(
+                variable.uname(),
+                variable.is_constant(),
```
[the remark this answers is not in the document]
martinjrobins: The spec says:
```A parameter’s value is set by its value attribute exactly once. If the parameter’s constant attribute (Sec26 tion 4.7.4) has the value “true”, then the value is fixed and cannot be changed except by an InitialAssignment.
27 These two methods of setting the parameter’s value differ in that the value attribute can only be used to set
28 it to a literal floating-point number, whereas InitialAssignment allows the value to be set using an arbitrary
29 mathematical expression (which, thanks to MathML’s expressiveness, may evaluate to a rational number)```

So I interpret that a parameter with `constant=true` can have an initialAssignment that is an expression (like `a+2`) that evaluates to a constant.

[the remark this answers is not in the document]
Comment (martinjrobins) [reply myokit/formats/sbml/_api.py:313]: Although there is a problem in my writer, in that it does not write out an initialAssignment rule for constant parameters with expressions!

File: myokit/formats/sbml/_api.py [hunk 14]
```diff
@@ -258,17 +348,21 @@ def add_compartment(self, sid):
...
 
-    def add_parameter(self, sid):
+    def add_parameter(self, sid, is_constant=True):
         """Adds a :class:`myokit.formats.sbml.Parameter` to this model."""
 
         sid = str(sid)
 
         self._register_sid(sid)
-        p = Parameter(self, sid)
+        p = Parameter(self, sid, is_constant)
```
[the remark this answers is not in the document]
martinjrobins: I guess there are both literal and constant parameters, but perhaps we can simplify the issue and only consider constant parameters. i.e. all parameter (even if constant) have an initialAssignment (and so should be added to _assignables), but in the case of a constant parameter that initialAssignment happens to evaluate to a constant (and the paramter has no rate rule)

myokit/formats/sbml/_api.py
```diff
@@ -572,6 +701,10 @@ def volume_units(self):
         'weber': myokit.units.Wb,
     }
```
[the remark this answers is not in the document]
martinjrobins: I'm happy either way, I did this to match `_base_units`, plus its only used internally

myokit/formats/sbml/_api.py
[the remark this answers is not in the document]
martinjrobins: actually, I've changed my mind, sbml defines two different ways to define a constant parameter so we probably should too:
a. Literal parameter: `constant=true` and `value=X`
b. Constant parameter set via initialAssigment: `constant=true` and no `value` set

myokit/tests/test_sbml_export.py
martinjrobins: hours are the last vestige of imperial units, the sooner we get rid of them the better!

myokit/formats/sbml/_api.py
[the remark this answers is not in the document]
martinjrobins: fixed this now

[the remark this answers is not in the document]
martinjrobins: A constant parameter now can be defined one of two ways:
1. Its a constant but not a literal, in this case `Qantity._initial_value` is set to the rhs expression (matching the way that sbml defines this type of parameter). `Quantity._value` is none.
2. Its a constant and a literal, this this case `Quantity._value` is set to the rhs expression, and `Quantity._initial_value` is None.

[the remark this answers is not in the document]
martinjrobins: let me know if this seems sensible @MichaelClerx

myokit/formats/sbml/_writer.py
martinjrobins: yea, I think this would be a premature optimisation, we can deal with it later if speed is an issue (I doubt it will be)

myokit/formats/sbml/_writer.py
martinjrobins: sure, I removed the `_` to mark this as public

myokit/formats/sbml/_api.py
```diff
@@ -514,6 +639,10 @@ def time_units(self):
         """Returns the default units for time, or dimensionless if not set."""
         return self._time_units
 
+    def units(self) -> Dict[str, myokit.Unit]:
+        """Returns a dict mapping sid to unit for all units in this model."""
+        return self._units
```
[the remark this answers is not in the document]
martinjrobins: I made this change, then reverted it since this would copy the dictionary each time (I think?)

myokit/formats/sbml/_writer.py
martinjrobins: I removed the `dict` constructor on this method so it should just return the reference to model._units

myokit/formats/sbml/_api.py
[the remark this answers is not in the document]
MichaelClerx: Yeah I was thinking that's what you'd want to do, to stop people from being able to modify internal properties? Units should be pretty fast to create so don't see this being a massive overhead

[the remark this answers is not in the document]
MichaelClerx: Just this one. Gimme a shout tomorrow morning and I'll merge it before noon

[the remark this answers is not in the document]
martinjrobins: ok yea, I see that you've gone for separate setters. Cool, I'll change this back

myokit/formats/sbml/_writer.py
martinjrobins: as above, I'll write the `has_units()`

MichaelClerx (approved):
Thanks so much Martin!

MichaelClerx:
🥳

PR #1116, opened 2025-01-10
bug: remove auto-register for Unit

### LINKED ISSUE #1115: [bug] the order of printing units can change the output to invalid mmt syntax
The following code produces units with invalid mmt syntax:

```python
unit1 = myokit.units.mol / (1e3 * myokit.units.g)
unit2 = 1e-12 * myokit.units.mol / (1e3 * myokit.units.g)
print(unit1)
print(unit2)
```

gives

```
[mol/g (0.001)]
[mol/g (0.001) (1e-12)]
```

The first call to `Unit.__str__` registers `mol/g (0.001)` as the preferred representation for this unit. Then the second call to `Unit.__str__` uses this registration and adds the multiplier difference, resulting in `[mol/g (0.001) (1e-12)]` which is invalid syntax

martinjrobins:
Fixes #1115

Removes auto-register in `Unit.__str__`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,12 @@ This page lists the main changes made to Myokit in each release.
 ## Unreleased
 - Added
 - Changed
+  - [#1116](https://github.com/myokit/myokit/pull/1116) Unit names passed to `Unit.register` are now validated using `myokit.check_name`, and the fourth argument to `Unit.register` is now called `preferred_representation`.
+  - [#1116](https://github.com/myokit/myokit/pull/1116) The method `Unit.register_preferred_representation` now checks that the given representation parses to the given unit.
 - Deprecated
 - Removed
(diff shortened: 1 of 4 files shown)


MichaelClerx:
- It wasn't the auto-register that was the problem: just that new representations were being built by adding multipliers to units with multipliers in their registration
- The existing tests didn't represent a valid use case, so were updated
- The issues with caching I was worried about were already solved, which is probably why I remembered to be worried :D

MichaelClerx:
@martinjrobins can you confirm that this fixes your use case?

martinjrobins (commented):
this works for me and looks great, thanks @MichaelClerx :)

MichaelClerx:
Just released [1.37.3](https://github.com/myokit/myokit/releases/tag/v1.37.3) for this

PR #1117, opened 2025-01-22
bug(sbml export): fix case of non-constant non-state variable

martinjrobins:
If a myokit model contains a non-constant variable that is not a state variable, the rule is added as an initialAssignment, should be an assigment rule.


MichaelClerx (approved):
Looks good to me

PR #1160, opened 2025-12-11
Heka time format fix for windows file

MichaelClerx:
Not sure equation is correct, waiting for more data to test against


MichaelClerx:
Both methods will back-date any times after 2040, whether we like it or not

PR #1164, opened 2026-01-14
Error handling and warnings in cvode 7+

MichaelClerx:
Two issues addressed here:

1. In cvode7+, the existing code added a 2nd error handler instead of replacing the default one
2. In cvode7+, it's possible to log warnings to a file (instead of stderr). Here writing to ""~, but not sure what that does. Probably safer to get a named temporary file on the python side and write to that instead. Or maybe ask on mailing list. Options seem to be set at compile time~


MichaelClerx:
https://github.com/llnl/sundials/issues/844

PR #1199, opened 2026-04-20
feat(diffsl): multiple protocols

martinjrobins:
This is a draft atm, I just had a question @MichaelClerx:

I'm currently implementing multiple protocol support for the diffsl export, to match the functionality available in Simulation and cause I need it in pkpd explorer :) 

My question is, I am primarily interested in instantaneous dosing events with duration 0. After looking around myokit I don't think your pacing system support this, is that right?

Would you be ok if I put a new arg on the diffsl exporter, called something like "delta_pace_threshold", so that if "duration" is less than that threshold, the pacing is treated as a delta function. This translates into a single reset event in diffsl, whereas an finite time duration event is two separate reset events, so more efficient....

### DIFF
--- a/myokit/formats/diffsl/_exporter.py
+++ b/myokit/formats/diffsl/_exporter.py
@@ -43,11 +43,13 @@ class DiffSLExporter(myokit.formats.Exporter):
         ``model``
             The :class:`myokit.Model` to export.
         ``protocol``
-            An optional :class:`myokit.Protocol` that defines a pacing or
-            dosing schedule.  When given, the exporter generates a hybrid ODE
-            model using DiffSL's ``N``, ``stop``, and ``reset`` constructs.
-            All events are expanded to one-off transitions; periodic events
-            require ``final_time`` to be set.
+            An optional :class:`myokit.Protocol` or sequence of
+            protocols that defines a pacing or dosing schedule. When given,
+            the exporter generates a hybrid ODE model using DiffSL's ``N``,
+            ``stop``, and ``reset`` constructs. A single protocol is expanded
+            into one-off transitions; multiple protocols are solved
+            sequentially for one window of ``final_time`` each, resetting the
+            state to the model's initial condition between windows.
         ``convert_units``
             If set to ``True`` (default), the method will attempt to convert to
             preferred units for voltage (mV), current (A/F), and time (ms).
@@ -61,12 +63,16 @@ class DiffSLExporter(myokit.formats.Exporter):
             be included in alphabetical order. All output variables must be
             time-varying (not constants).
         ``final_time``
-            Required when ``protocol`` is provided. Events are expanded up to
-            (but not including) this time. Must be a finite positive number.
-            Ignored when ``protocol`` is ``None``.
+            Required when ``protocol`` is provided. For a single protocol,
+            events are expanded up to (but not including) this time. For a
+            sequence of protocols, each protocol is expanded over
+            ``[0, final_time)`` before being offset into its own window. Must
+            be a finite positive number. Ignored when ``protocol`` is
+            ``None``.
         """
         # Validate protocol / final_time combination
-        if protocol is not None:
+        protocols = self._normalize_protocols(protocol)
+        if protocols is not None:
             if final_time is None:
                 raise myokit.ExportError(
                     'A final_time must be provided when exporting with a '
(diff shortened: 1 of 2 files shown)


martinjrobins:
hmmm, after thinking about this some more, this would be difficult to do in general, you would have to know how to integrate the RHS with pace as a delta. In my case its easy as pace is just added to the RHS, but in general this is probably too complicated to bother with. Nevermind!

martinjrobins:
ok, I've concluded this is not what I want to do. It should be a dict of protocols pacing different variables, not a list of protocols pacing different simulations. Starting again.....

MichaelClerx:
> My question is, I am primarily interested in instantaneous dosing events with duration 0. After looking around myokit I don't think your pacing system support this, is that right?

So is the plan that a particular type of dosing is a certain variable? And you ignore the duration for those? Or something with edge up and edge down, stuff like that? Just thinking what we can do to make this easier

martinjrobins:
Yea, at the moment we just put in a manual pulse with a very short duration, but this introduces an artificial short timescale and slows down the solver. I was hoping to use the export to diffsl to encode this as a reset, but it only works if the dose is an additive term in the equations, and you can't guarentee this.

MichaelClerx:
It should be quite simple to update the event protocol implementations to allow zero-duration events. A few knock-on effects though which are harder to predict

- There may be more implementation than you think, 1 Python, 1 C, but also e.g. in exports
- We'd need some guards to stop `next` in a protocol causing infinite loops or other weirdness

martinjrobins:
but how do you actually go about integrating the delta function? that is mainly what stopped me. 

Say if you had:
dydt = y * p^2, or dydt = y / (2 - p)

where p is bound to some protocol. 

It only really makes sense for these two situations:
dydt = f(y, t) p
and
dydy = f(y, t) + p

martinjrobins:
I guess you could raise an error if you have anything else...

mirams:
It feels as if this kind of thing is better done with 'events' (altering state variables at given times or when thresholds are reached) than trying to integrate delta functions which will always be numerically disasterous I would think.

martinjrobins:
yea, diffsl uses a dedicated stop (i.e. root crossing) tensor that encode *when* the event is, and a reset tensor that encodes *what* the event does. see https://martinjrobins.github.io/diffsl/odes.html#stopping-and-resetting-the-ode-system

MichaelClerx:
Fair enough. I don't think you want to be integrating a kronecker delta 😄 
And I don't think you want to deal with the ramifications of adding that type of reset to Myokit either

The simple solutions, if you wanted to _implement_ this behaviour in Myokit would be:
1. Run the simulation until the time of the event, make the change, run again (e.g. s.run(1), s.set_constant([might require an s.get_value() or something here if it isn't a state or logged variable), s.run(1))
2. For dosing, not bouncing balls, stop defining the dosing event as an impulse, but instead have a `dose = f() + added_dosing`, `dot(added_dosing) = pace * ddose_dt` where duration and `ddose_dt` are chosen so that the integral over the duration of the dosing event equals the added dose

But neither are satisfactory for the use case of defining the model in Myokit then exporting to DiffSL, I guess?

MichaelClerx:
I'm veeeeery hesitant to add resets to Myokit, because they are the worst part of both SBML and CellML, and I suspect (in a hand-wavy and slightly facetious way) that this is because it amounts to encoding a procedure in a declarative language, rather than a procedural one

martinjrobins:
resets are a pain to add I agree, particularly for implementing sensitivities and adjoints correctly. So I fully support your reluctance! For the moment I'll just go with option 2 above and using a more beefy server, that is what we've been anyway. 

You can implement events in a declaritive way though, it doesn't have to be procedural, although its often implemented as such. The equations that I use are:

`y(t_0)^+ = r(y(t_0)^-), where 0 = s(y(t_0))`, so you just define the functions `r` and `s`  along with your normal `f` function for the RHS and solve.

martinjrobins:
then, if you want to have a sequence of events you need to have a set of functions indexed by say `N`, so `f_N`, `r_N` & `s_N`, and then you need to define a way to transition between different `N` (I use a vector valued `s` function and the particular element that crosses 0 defines the next value of `N`)

martinjrobins:
so yes, it can get a bit convoluted ..... ;)

PR #1200, opened 2026-04-21
feat(diffsl): support protocol dict

martinjrobins:
- supports a dictionary of { binding_name -> Protocol } for the protocol arg for diffsl export
- plus misc fixes based on testing with pkpd explorer


docs/source/api_formats/diffsl.rst
```diff
@@ -11,6 +11,12 @@ Myokit provides export of models to the `DiffSL language <https://martinjrobins.
 **Protocol export** generates a *hybrid* ODE model using DiffSL's ``N`` and
 ``stop`` constructs, see the `DiffSL hybrid ODE documentation <https://martinjrobins.github.io/diffsl/odes.html>`_.
 
+The ``protocol`` argument can be either a single :class:`myokit.Protocol`
+for the ``pace`` binding or a ``dict`` mapping binding names to protocols.
+When multiple bindings are provided, Myokit expands them to a shared phase
```
MichaelClerx: I don't know what a shared phase timeline is, but makes sense in DiffSL context?

myokit/formats/diffsl/_exporter.py
```diff
@@ -65,8 +67,12 @@ def model(self, path, model, protocol=None, convert_units=True,
             (but not including) this time. Must be a finite positive number.
             Ignored when ``protocol`` is ``None``.
         """
+        # Normalize protocols to a binding -> Protocol mapping
+        protocols = self._normalize_protocols(protocol)
+        protocol_bindings = set(protocols)
```
MichaelClerx: You can just use `protocols`' keys for this

myokit/formats/diffsl/_exporter.py
```diff
@@ -88,8 +94,16 @@ def model(self, path, model, protocol=None, convert_units=True,
                 'DiffSL export requires a valid model.'
             ) from e
 
+        # Validate any requested protocol bindings against the model
+        for label in protocol_bindings:
+            if model.binding(label) is None:
+                raise myokit.ExportError(
```
MichaelClerx: This needs a bit of clarification in the export docstring, and possibly in the error message. The standard Myokit behaviour is to allow this

myokit/formats/diffsl/_exporter.py
```diff
@@ -434,43 +519,52 @@ def _generate_diffsl(self, model, inputs, outputs, phases=None):
...
             export_lines.append('')
 
         return '\n'.join(export_lines)
 
+    def _normalize_protocols(self, protocol):
+        """
+        Normalize a protocol specification to an ordered binding map.
+        """
+        if protocol is None:
+            return collections.OrderedDict()
```
MichaelClerx, line 536: We're not supporting 2 anymore so an ordinary dict is the same?

myokit/formats/diffsl/_exporter.py
```diff
@@ -88,8 +94,16 @@ def model(self, path, model, protocol=None, convert_units=True,
...
+        # Validate any requested protocol bindings against the model
+        for label in protocol_bindings:
+            if model.binding(label) is None:
+                raise myokit.ExportError(
+                    f'No variable found with binding "{label}".'
+                )
+
         # Store qualified names of inputs/outputs before model prep
         # and validate that they are appropriate for DiffSL export
+        allowed_bound_inputs = protocol_bindings if protocols else {'pace'}
```
MichaelClerx: Has to be a dict after normalisation?

MichaelClerx: Default Myokit behaviour w.r.t. bindings is that they have a fallback RHS when a binding is specified in the model but not used. So an alternative logic here would be that _any_ binding is OK for an input (the  model should be written that way). But up to you!

MichaelClerx: I wouldn't make 'pace' an exception if 'pace' is not guaranteed to be something the DiffSL solver uses though. So `if x in allowed_bound_inputs` should be the same as asking `if x in protocols` again

MichaelClerx (changes requested):
Hey @martinjrobins ,
 some early questions about this one so passing it back to you before I dig in

docs/source/api_formats/diffsl.rst
martinjrobins: more from the variable names used in the code, I'll try and tidy this up

myokit/formats/diffsl/_exporter.py
martinjrobins: yes, protocols will be a dict here so testing if its empty.

Regarding the pace exception, this is to make it backwards compatible with the previous behaviour if protocols is not provided, I'm unsure if I should keep this behaviour, @kwabenantim?

martinjrobins: note `allowed_bound_inputs` is only really used to check that you get the right error message below "Input variable... cannot be both an input and be driven by protocol..."

myokit/formats/diffsl/_exporter.py
martinjrobins: actually, if myokit allows this I'll just remove this error

myokit/formats/diffsl/_exporter.py
martinjrobins: "Default Myokit behaviour w.r.t. bindings is that they have a fallback RHS when a binding is specified in the model but not used".

Yea, this should be what is happening, if a variable has a binding that is not provided, it just uses its normal rhs

myokit/formats/diffsl/_exporter.py
martinjrobins: Sorry for the confusion. My previous PR (which I closed) was implementing multiple protocols on the *same* binding, this one is just implementing a single protocol for *multiple* bindings, hence there is more than two entries in the dict. It must be ordered because diffsl requires an ordering (i.e. which binding goes in which index of the vector)

myokit/formats/diffsl/_exporter.py
kwabenantim: > Regarding the pace exception, this is to make it backwards compatible with the previous behaviour if protocols is not provided

I think it's okay to remove the old behaviour if it doesn't generalise well enough. We can probably use something like `exporter.model(protocol={'pace': Protocol()}, ... )` to keep the old behaviour if needed.

docs/source/api_formats/diffsl.rst
```diff
@@ -11,6 +11,14 @@ Myokit provides export of models to the `DiffSL language <https://martinjrobins.
 **Protocol export** generates a *hybrid* ODE model using DiffSL's ``N`` and
 ``stop`` constructs, see the `DiffSL hybrid ODE documentation <https://martinjrobins.github.io/diffsl/odes.html>`_.
 
+The ``protocol`` argument can be either a single :class:`myokit.Protocol`
+for the ``pace`` binding or a ``dict`` mapping binding names to protocols.
+Protocol entries for bindings that are not present in the model are ignored.
+When multiple bindings are provided, Myokit expands them to a shared
```
MichaelClerx:
```suggestion
When multiple bindings are provided, the exporter expands them to a shared
```

because what comes next is all about diffsl, really

myokit/formats/diffsl/_exporter.py
```diff
@@ -43,11 +52,15 @@ def model(self, path, model, protocol=None, convert_units=True,
...
         ``protocol``
-            An optional :class:`myokit.Protocol` that defines a pacing or
-            dosing schedule.  When given, the exporter generates a hybrid ODE
-            model using DiffSL's ``N``, ``stop``, and ``reset`` constructs.
-            All events are expanded to one-off transitions; periodic events
-            require ``final_time`` to be set.
+            An optional :class:`myokit.Protocol` or ``dict`` mapping binding
+            names to :class:`myokit.Protocol` objects that define pacing or
+            dosing schedules. If a map is not given then the binding name is
+            assumed to be `pace`. When given, the exporter generates a
```
MichaelClerx:
```suggestion
            assumed to be ``pace``. When given, the exporter generates a
```

myokit/formats/diffsl/_exporter.py
```diff
@@ -65,29 +78,33 @@ def model(self, path, model, protocol=None, convert_units=True,
             (but not including) this time. Must be a finite positive number.
             Ignored when ``protocol`` is ``None``.
         """
+        # Check model validity
+        try:
+            model.validate()
+        except myokit.MyokitError as e:
+            raise myokit.ExportError(
+                "DiffSL export requires a valid model."
```
MichaelClerx: Style-wise Myokit usually uses ' rather than "

Just cause I had massive RSI at some point and was avoiding shift keys. Feel free to ignore :D

myokit/formats/diffsl/_exporter.py
MichaelClerx: But is `dict` not ordered since Python 3?

MichaelClerx (dismissed):
Looks good to me, once coverage is fixed!

myokit/formats/diffsl/_exporter.py
martinjrobins: this is what autoformatters are made for... :) I'll swap for '

myokit/formats/diffsl/_exporter.py
martinjrobins: yea, but I'd just prefer to be explicit about it, I can never remember which version of which language implements an ordered vs unordered map.....

myokit/formats/diffsl/_exporter.py
martinjrobins: @MichaelClerx I added single quote style to the ruff settings in pyproject.toml and reformated this file to use single quote

MichaelClerx: Nice! Do you use ruff for type checking ?

martinjrobins: yea, it has a variety of linter checks as well as type checking. You get the autofomatting for free :)

PR #1201, opened 2026-07-20
feat(diffsl): support piecewise function

martinjrobins:
support the new `piecewise` function in diffsl: https://martinjrobins.github.io/diffsl/functions.html

### DIFF
--- a/myokit/formats/diffsl/__init__.py
+++ b/myokit/formats/diffsl/__init__.py
@@ -50,6 +50,7 @@ keywords = [
     'log',
     'M',
     'out',
+    'piecewise',
     'pow',
     'sigmoid',
     'sin',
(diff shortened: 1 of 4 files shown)


martinjrobins:
few CI errors, but they don't seem like anything to do with this PR. UPDATE: fixed them, just some very small complex values sneaking in

MichaelClerx (approved):
Looks good!

Are the rules for half-open intervals etc the same for both languages, or do we need to at least make mention?

martinjrobins:
> Are the rules for half-open intervals etc the same for both languages, or do we need to at least make mention?

the translation from myokit piecewise to diffsl piecewise should preserve existing semantics. The myokit piecewise conditions are true/false, but since diffsl doesn't have boolean or conditional statements it uses a float x that is "true" if x >=0 and "false" if x < 0 (ie. heaviside function). The first branch that evaluates to "true" is taken, just like myokit's piecewise

kwabenantim (approved):
This looks good to me too!

martinjrobins:
@MichaelClerx should I merge this in?

